PR #55, opened 2024-07-04
Add integration tests

eminano:
This PR adds integration tests to the stream package, using testcontainers to spawn the relevant resources to run the tests against. They are controlled by an environment variable (`PGSTREAM_INTEGRATION_TESTS`), which if not set will not run them. 

There are two independent make targets, and two CI actions to run unit tests and integration tests separately.


pkg/stream/stream_start.go
```diff
@@ -50,6 +49,7 @@ func Start(ctx context.Context, logger loglib.Logger, config *Config, meter metr
 		if err != nil {
 			return fmt.Errorf("error setting up postgres replication handler")
 		}
+		defer replicationHandler.Close()
```
exekias, line 52: can we really close this at the end of start? Would it mean `Run` is a better name?

exekias (approved):

pkg/stream/stream_start.go
eminano: `Start` is a blocking call, when it exits it cleans up all started resources. That's where all others are defered closed/stopped. I can change it to `Run` if that's clearer though 👍 .

exekias (approved):

PR #59, opened 2024-08-05
Support PEM certificates as well as files

eminano:
Expand support of TLS library to accept a certificate file or the certificate PEM content directly.


pkg/tls/tls.go
```diff
@@ -72,12 +80,23 @@ func getCertificates(clientCertFile, clientKeyFile string) ([]tls.Certificate, e
...
-	defer file.Close()
 
-	return io.ReadAll(file)
+	return os.ReadFile(cert)
+}
+
+// isPEMString returns true if the provided string match a PEM formatted certificate.
+func isPEMString(s string) bool {
+	// trim the certificates to make sure we tolerate any yaml weirdness
+	trimmedStr := strings.TrimSpace(s)
```
ryanslade, line 96: nit: You could reuse `s` here.

```suggestion
	s = strings.TrimSpace(s)
```

pkg/tls/tls.go
```diff
@@ -72,12 +80,23 @@ func getCertificates(clientCertFile, clientKeyFile string) ([]tls.Certificate, e
...
+
+	// we assume that the string starts with "-" and let further validation
+	// verifies the PEM format. When migrating from pkcs12 to PEM a "Bag
```
ryanslade, line 99:
```suggestion
	// verify the PEM format. When migrating from pkcs12 to PEM a "Bag
```

ryanslade (dismissed):
Code looks good. It is a bit confusing to overload the config to be a path or cert content though. Is it possible to update the config itself to have different properties and it's expected that the client only provides one or the other?

pkg/tls/tls.go
eminano: I agree, it's a bit confusing. I added this as it's a supported functionality in the xata codebase, so it would otherwise have trouble connecting with the given config. I can update it so that it's different methods to be more explicit about what's being passed (originally this library only supported files for example).

ryanslade (approved):
Not a deal breaker, but might be worth adding some tests?

PR #60, opened 2024-08-05
Update lib hooks

eminano:
This PR updates the hooks that allow to customise the use for the pgstream library:
- Add support to use a custom search index name adapter and a mapper
- Add granularity to filter wal data and schema events separately

Small improvements included:
- Update of pgx dependency 
- Removal of panic during log entry unmarshaling
- Apply search store retry logic by default (if no config is provided, there's a default that applies). If no retries are required, then the config must be explicitly updated accordingly.

### DIFF
--- a/pkg/schemalog/log_entry.go
+++ b/pkg/schemalog/log_entry.go
@@ -64,8 +64,6 @@ func (m *LogEntry) UnmarshalJSON(b []byte) error {
 			if err := json.Unmarshal([]byte(schemaStr), &m.Schema); err != nil {
 				return err
 			}
-		default:
-			panic(fmt.Sprintf("unmarshal LogEntry, got unexpected key when unmarshalling: %s", k))
 		}
 	}
 
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -169,7 +169,7 @@ func parseSearchProcessorConfig() *stream.SearchProcessorConfig {
 		Store: opensearch.Config{
 			URL: searchStore,
 		},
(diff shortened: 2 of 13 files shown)


pkg/schemalog/log_entry.go
```diff
@@ -64,8 +64,6 @@ func (m *LogEntry) UnmarshalJSON(b []byte) error {
 			if err := json.Unmarshal([]byte(schemaStr), &m.Schema); err != nil {
 				return err
 			}
-		default:
```
ryanslade, line 67: Why do we need the custom `UnmarshalJSON` method? It doesn't seem to be doing anything special apart from the `schema` case where it checks the length. If this is the only special case then it could be moved into some code on the `Schema` type.

eminano: To be honest, I'm not sure why we needed this. I will have a look and see if we can get rid of it (will open a separate PR for that if needed) 👍

eminano: I was looking into this, we need the custom unmarshal because we use it to unmarshal from the WAL event, from a map[string]any representation instead of a log entry json one. 
https://github.com/xataio/pgstream/blob/main/pkg/wal/processor/wal_processor.go#L51-L53

ryanslade: If that's the only place then it would probably be better to do the conversion in `WalDataToLogEntry` by looking at the column name / values and populating the LogEntry there rather than having that logic buried in the JSON decoder.

Not urgent, just feels like a bit of tech debt.

PR #66, opened 2024-08-29
Improve support for elasticsearch replication

eminano:
This PR improves the support for elasticsearch replication, as well as improving the feature support for opensearch replication. Before this PR we were using a go-elasticsearch library version that supported opensearch (before it introduced version/client/server checks in 2021). This prevented any updates, and it had limitations when using vector fields with elasticsearch.

By separating the opensearch and elasticsearch clients we can address the slight differences for each in the index, mapping and overall library use. The separation is done at the library/pkg level, the rest of changes are meant to align with a decoupled approach, avoiding leaking implementation details. 

Main changes:
- Addition of elasticsearch client and creation of an abstracted search store library with the shared functionality, used as an access point for the rest of the code, abstracting the implementation details.
- Addition of a search store mapper, along with the opensearch and elasticsearch implementations, that takes care of the specificities of each regarding the mapping (ex., [`knn_vector` vs `dense_vector`](https://opster.com/guides/elasticsearch/machine-learning/vector-search-in-opensearch-vs-elasticsearch/)) .
- Update search batch processor to use the newly abstracted library
- Update configuration to be able to provide an elasticsearch or opensearch store url 
- Addition of integration tests to cover the postgres to elasticsearch integration


pkg/wal/processor/search/store/search_store.go
```diff
@@ -9,63 +9,75 @@ import (
...
-	os, err := es.NewClient(cfg.URL)
+	var searchStore searchstore.Client
+	var err error
+	switch {
+	case cfg.OpenSearchURL != "":
+		searchStore, err = opensearchstore.NewClient(cfg.OpenSearchURL)
+	case cfg.ElasticsearchURL != "":
+		searchStore, err = elasticsearchstore.NewClient(cfg.ElasticsearchURL)
+	default:
+		return nil, errors.New("invalid search store configuration provided")
```
exekias, line 55: it would probably be good to provide more detail in all the different error cases here:

* Both OpenSearch + Elasticsearch are configured ❗ 
* None of them is ❗

exekias (approved):
LGTM besides user input handling!

PR #78, opened 2024-10-01
Allow to define a version finder that uses LSN

eminano:
Update version finder to support using the LSN instead of a column.

### DIFF
--- a/pkg/wal/processor/translator/wal_translator.go
+++ b/pkg/wal/processor/translator/wal_translator.go
@@ -29,7 +29,7 @@ type Translator struct {
 	skipSchemaEvent      schemaEventFilter
 	schemaLogStore       schemalog.Store
 	idFinder             columnFinder
-	versionFinder        columnFinder
+	versionFinder        columnFinderWithErr
 }
 
 type walToLogEntryAdapter func(*wal.Data) (*schemalog.LogEntry, error)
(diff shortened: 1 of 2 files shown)


pkg/wal/processor/translator/wal_translator.go
```diff
@@ -231,10 +234,18 @@ func (t *Translator) fillEventMetadata(event *wal.Data, log *schemalog.LogEntry,
 			continue
 		}
 
-		if t.versionFinder != nil && t.versionFinder(col, tbl) && !foundVersion {
-			foundVersion = true
-			event.Metadata.InternalColVersion = col.PgstreamID
-			continue
+		if t.versionFinder != nil && !foundVersion {
```
ryanslade, line 237: Instead of checking in a few places if `versionFinder` is nil, can we not return early at the beginning if it is nil and remove the checks?

ryanslade: Ah, nevermind, I see you still need to check by id too.

eminano: Yeah, the version finder is optional, but we reuse the same loop for finding the identity as well, which is required, to avoid having two loops over the table columns. 
The version finder is a hook to introduce custom functionality when identifying the version column. In the case of xata, we have our own logic to identify the version, and this update helps the transition towards using the LSN instead, with feature flag granularity.

ryanslade (approved):

pkg/wal/processor/translator/wal_translator.go
eminano: We also don't want to default to using the LSN if a version finder is provided and no version is found, since that would mean events would be replicated using the LSN, whereas it should be treated as an invalid event from the user point of view. This is why we only check if the version is found when the version finder is provided (hence the second check on the versionFinder you mentioned).

PR #79, opened 2024-10-02
Add snapshot store

eminano:
This PR adds a snapshot store to persist snapshot requests. This will be used by a snapshot server which will receive the requests, in order to have a persisted list of snapshots and their status.

Part of https://github.com/xataio/pgstream/issues/80

### DIFF
--- a/pkg/snapshot/snapshot.go
+++ b/pkg/snapshot/snapshot.go
@@ -1,0 +1,32 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package snapshot
+
+type Snapshot struct {
+	SchemaName          string
+	TableName           string
+	IdentityColumnNames []string
+	Status              Status
+	Error               error
+}
+
+type Status string
+
+const (
+	StatusRequested = Status("requested")
+	StatusStarted   = Status("started")
+	StatusCompleted = Status("completed")
+)
+
+func (s *Snapshot) IsValid() bool {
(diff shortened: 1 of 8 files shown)


pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,32 @@
...
+	IdentityColumnNames []string
+	Status              Status
+	Error               error
+}
+
+type Status string
+
+const (
+	StatusRequested = Status("requested")
+	StatusStarted   = Status("started")
```
exekias, line 17: nit: in progress?

exekias (approved):

pkg/snapshot/store/postgres/pg_snapshot_store.go
```diff
@@ -0,0 +1,116 @@
...
+	_, err = s.conn.Exec(ctx, uniqueIndexQuery)
+	if err != nil {
+		return fmt.Errorf("error creating unique index on snapshots postgres table: %w", err)
+	}
+
+	return err
+}
+
+func snapshotsTable() string {
+	return fmt.Sprintf("%s.%s", store.SchemaName, store.TableName)
```
exekias, line 115: we should use `pq.QuoteIdentifier`, especially if users can modify these params

eminano: I don't think users can modify these parameters, but using quote identifier anyway.

PR #83, opened 2024-10-24
Delete opensearch schema synchronously

### LINKED ISSUE #77: [Search Processor] Index not found after schema dropped
Currently the search batch indexer has a background process that registers index delete requests and processes them asynchronously. This can cause issues if the delete happens after an index was recreated, causing data loss for subsequent documents.

eminano:
This PR simplifies the search index delete by handling it synchronously instead of asynchronously. 
With the current async behaviour, it was possible to delete an index after it had been recreated, causing "index not found" errors when subsequent data updates were received. By making it synchronous(with retries), we will prevent any further events from being applied until it completes, ensuring the delete happens before any subsequent recreation happens.

Closes #77


exekias (approved):
nice!

PR #87, opened 2024-11-18
Reinforce kafka msg size guards

eminano:
Currently there's a guard in place to make sure there are no messages sent to kafka that go over the limit in size. However, the check is only accounting for the serialised bytes of the value in the message, but not for the rest of the message (headers, metadata...). This PR updates the size check to ensure we leave some buffer to account for those extra parts of the message, discarding the message and logging it as data loss if the value is more than 95% of the limit.

An error log is added to identify when the send thread stops unexpectedly due to kafka errors.

### DIFF
--- a/pkg/wal/processor/kafka/wal_kafka_batch_writer.go
+++ b/pkg/wal/processor/kafka/wal_kafka_batch_writer.go
@@ -138,10 +138,11 @@ func (w *BatchWriter) ProcessWALEvent(ctx context.Context, walEvent *wal.Event) 
 		if err != nil {
 			return fmt.Errorf("marshalling event: %w", err)
 		}
-		// check if walEventBytes is larger than the Kafka accepted max message size
-		if len(walDataBytes) > int(w.maxBatchBytes) {
+		// check if walEventBytes is larger than 95% of the Kafka accepted max
+		// message size to allow for some buffer for the rest of the message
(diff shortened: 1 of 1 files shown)


pkg/wal/processor/kafka/wal_kafka_batch_writer.go
```diff
@@ -138,10 +138,11 @@ func (w *BatchWriter) ProcessWALEvent(ctx context.Context, walEvent *wal.Event)
 		if err != nil {
 			return fmt.Errorf("marshalling event: %w", err)
 		}
-		// check if walEventBytes is larger than the Kafka accepted max message size
-		if len(walDataBytes) > int(w.maxBatchBytes) {
+		// check if walEventBytes is larger than 95% of the Kafka accepted max
+		// message size to allow for some buffer for the rest of the message
+		if len(walDataBytes) > int(0.95*float64(w.maxBatchBytes)) {
```
exekias, line 143: this is probably good enough, although it can be a lot of buffer for high max bytes settings, I wonder if there is a predictable payload size math (ie 100bytes padding)

exekias (approved):
I believe we can merge this as it is, left a comment on the approach, but I don't have a better one :)

pkg/wal/processor/kafka/wal_kafka_batch_writer.go
eminano: I'll have a think/look.

PR #88, opened 2024-11-26
Add code coverage to github actions

eminano:


exekias:
sorry I see it now, you will need this step for things to work: https://github.com/xataio/maki/blob/main/.github/workflows/test.yml#L33-L37

I wonder what will be the deal with 2 different tests going (unit + integration)

PR #100, opened 2024-12-10
Add support to provide LSN

eminano:
This PR updates the replication handler to support retrieving the current LSN as well as starting replication from a provided LSN. This will be used by the snapshoting behaviour, to ensure we can retrieve the LSN before getting the transaction snapshot, so that when we finish the snapshot we can resume/start replication from the last processed event.

### DIFF
--- a/pkg/wal/replication/instrumentation/instrumented_replication_handler.go
+++ b/pkg/wal/replication/instrumentation/instrumented_replication_handler.go
@@ -47,6 +47,10 @@ func (h *Handler) StartReplication(ctx context.Context) error {
 	return h.inner.StartReplication(ctx)
 }
 
+func (h *Handler) StartReplicationFromLSN(ctx context.Context, lsn replication.LSN) error {
+	return h.inner.StartReplicationFromLSN(ctx, lsn)
+}
+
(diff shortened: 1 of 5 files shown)


eminano:
> ### Merging this branch changes the coverage (1 decrease, 1 increase)
> Impacted Packages 	Coverage Δ 	🤖
> github.com/xataio/pgstream/pkg/wal/replication 	0.00% (ø) 	
> github.com/xataio/pgstream/pkg/wal/replication/instrumentation 	0.00% (ø) 	
> github.com/xataio/pgstream/pkg/wal/replication/mocks 	19.64% (**+0.08%**) 	👍
> github.com/xataio/pgstream/pkg/wal/replication/postgres 	76.14% (**-4.88%**) 	👎
> Coverage by file

The decrease in coverage for the postgres package is caused by some of the initialisation being moved to the constructor, instead of the start replication method (this allows us to identify quicker if there's an issue with replication and exit early rather than waiting for the process to start replicating to flag the issue).

The constructor contains postgres implementation specific logic that is not meant to be covered by unit tests.

exekias (approved):

PR #102, opened 2024-12-10
Add snapshot generator recorder

eminano:
This PR adds a decorator around a snapshot generator to record the request activity. It checks if there are any existing completed snapshots for the snapshot request, and parses the result to make sure if there are existing snapshots that completed with errors, only the failed tables are retried. 

It helps keep track of when a snapshot was requested and completed, and any potential errors during the processing. This will prevent the configured initial snapshot to be re-run every time pgstream is started.

### DIFF
--- a/pkg/snapshot/generator/snapshot_generator_recorder.go
+++ b/pkg/snapshot/generator/snapshot_generator_recorder.go
@@ -11,6 +11,8 @@ import (
 	snapshotstore "github.com/xataio/pgstream/pkg/snapshot/store"
 )
 
+// SnapshotRecorder is a decorator around a snapshot generator that will record
+// the snapshot request status.
 type SnapshotRecorder struct {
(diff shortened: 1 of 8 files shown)


pkg/snapshot/generator/snapshot_generator_recorder.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+type SnapshotRecorder struct {
+	wrapped SnapshotGenerator
+	store   snapshotstore.Store
+}
+
+const updateTimeout = time.Minute
+
+func NewSnapshotRecorder(store snapshotstore.Store, generator SnapshotGenerator) *SnapshotRecorder {
```
exekias, line 21: worth adding a comment explaining this object

exekias (approved):

PR #124, opened 2025-01-24
Rename translator to injector

eminano:
This PR renames the processor `translator` decorator with a more relevant name, `injector`. This layer is in charge of injecting pgstream metadata into the wal event, rather than translating.


exekias (approved):

agedemenli (approved):
Much better <3

PR #140, opened 2025-02-12
Add transformers package

eminano:
This PR adds a transformers package which defines the transformer interface and some implementations for string and first name integrating with opensource implementations ([neosync](https://github.com/nucleuscloud/neosync), [greenmask](https://github.com/eminano/greenmask)).


agedemenli (approved):
This looks great

PR #143, opened 2025-02-14
Add transformer processor

eminano:
This PR adds a wal processor decorator that will apply the configured column transformers to the incoming wal events.
The rules are defined as a yaml file to keep the format flexible and extensible, as well as provide easier integration with external transformers.


README.md
```diff
@@ -234,6 +235,15 @@ One of exponential/constant backoff policies can be provided for the search stor
 
 </details>
 
+<details>
+  <summary>Transformer</summary>
+
+| Environment Variable            | Default | Required | Description                                                          |
+| ------------------------------- | ------- | -------- | -------------------------------------------------------------------- |
+| PGSTREAM_TRANSFORMER_RULES_FILE | N/A     | No       | Filepath pointing to the yaml file containing the transformer rules. |
```
exekias, line 243: I wonder if eventually we will want all settings in a single file, but this we can figure out later

pkg/wal/processor/transformer/wal_transformer.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+type Option func(t *Transformer)
+
+// New will return a transformer processor wrapper that will transform incoming
+// wal event column values as configured by the transformation rules.
+func New(cfg *Config, processor processor.Processor, opts ...Option) (*Transformer, error) {
+	rules, err := readRulesFromFile(cfg.TransformerRulesFile)
+	if err != nil {
+		return nil, err
```
exekias, line 37: maybe we want to add some context to the error here

exekias (approved):

pkg/wal/processor/transformer/wal_transformer.go
eminano: The `readRulesFromFile` function returns already context within the errors:
https://github.com/xataio/pgstream/blob/add-transformer-processor/pkg/wal/processor/transformer/wal_transformer_rules.go#L30-L42

PR #154, opened 2025-02-20
Add gm integer transformer

agedemenli:
This PR introduces `IntegerTransformer` which uses `greenmasktransformers.RandomInt64Transformer` under the hood.

### DIFF
--- a/pkg/transformers/greenmask/greenmask_integer_transformer.go
+++ b/pkg/transformers/greenmask/greenmask_integer_transformer.go
@@ -1,0 +1,94 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package greenmask
+
+import (
+	"encoding/binary"
+	"fmt"
+
+	greenmasktransformers "github.com/eminano/greenmask/pkg/generators/transformers"
+	"github.com/xataio/pgstream/pkg/transformers"
+)
+
+const DefaultSize = 4
+
+type IntegerTransformer struct {
+	transformer *greenmasktransformers.RandomInt64Transformer
+}
+
+func NewIntegerTransformer(generator transformers.GeneratorType, params transformers.Parameters) (*IntegerTransformer, error) {
+	size, err := findParameter(params, "size", int(DefaultSize))
+	if err != nil {
+		return nil, fmt.Errorf("greenmask_integer: size must be an integer: %w", err)
+	}
+	if size < 1 {
+		return nil, fmt.Errorf("greenmask_integer: size must be greater than 0")
+	}
+	if size > 8 {
+		return nil, fmt.Errorf("greenmask_integer: size must be less than or equal to 8")
+	}
+
+	DefaultMinValue := minValueForSize(size)
+	DefaultMaxValue := maxValueForSize(size)
+
+	min_value, err := findParameter(params, "min_value", DefaultMinValue)
+	if err != nil {
+		return nil, fmt.Errorf("greenmask_integer: min_value must be an integer: %w", err)
+	}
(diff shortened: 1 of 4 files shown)


pkg/transformers/greenmask/greenmask_integer_transformer_test.go
```diff
@@ -0,0 +1,415 @@
...
+package greenmask
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/xataio/pgstream/pkg/transformers"
+)
+
+func TestNewIntegerTransformer(t *testing.T) {
```
eminano, line 12: In most cases, unit tests can be run in parallel, since they don't affect each other. This helps speeding up the test runs. 
You just need to add a `t.Parallel()` both here and below in the for loop, so that this test, and all the subcases are run concurrently. 
You can see the examples in the other test files:
https://github.com/xataio/pgstream/blob/main/pkg/transformers/greenmask/greenmask_string_transformer_test.go#L13
https://github.com/xataio/pgstream/blob/main/pkg/transformers/greenmask/greenmask_string_transformer_test.go#L82

pkg/transformers/greenmask/greenmask_integer_transformer_test.go
```diff
@@ -0,0 +1,415 @@
...
+	tests := []struct {
+		name        string
+		generator   transformers.GeneratorType
+		params      transformers.Parameters
+		wantErr     bool
```
eminano, line 17: It's usually easier to have a `wantErr` of type `error` (instead of a boolean and a string) that you can then check and validate against nil and not nil values by using the golang [`errors`](https://pkg.go.dev/errors) utilities, like `errors.ErrorIs`. It simplifies the test cases, and also keeps the logic for validation below simple.
`errors.ErrorIs` is often also used throughout the code to catch errors of a particular type from different packages without having to know the specifics of the error message. 

In this case, you can check that the error returned fits the invalid parameters type and simplify the test body. Example: https://github.com/xataio/pgstream/blob/main/pkg/transformers/greenmask/greenmask_string_transformer_test.go#L69

pkg/transformers/greenmask/greenmask_integer_transformer_test.go
```diff
@@ -0,0 +1,415 @@
...
+	}
+}
+
+func TestIntegerTransformer_Transform_Random(t *testing.T) {
+	transformer, err := NewIntegerTransformer(transformers.Random, transformers.Parameters{
+		"size":      8,
+		"min_value": int64(-100),
+		"max_value": int64(100),
+	})
+	if err != nil {
```
eminano, line 123: We use the [`testify`](https://github.com/stretchr/testify) go library, which simplifies all these very verbose test validations. You can use `require`, there are some examples here:
https://github.com/xataio/pgstream/blob/main/pkg/transformers/greenmask/greenmask_string_transformer_test.go

eminano (commented):
Nice work! The comments are mostly aimed at giving some guidance around go and keeping the pgstream project consistent.

pkg/transformers/greenmask/greenmask_integer_transformer.go
```diff
@@ -0,0 +1,115 @@
...
+
+const (
+	defaultSize = 4
+)
+
+type IntegerTransformer struct {
+	transformer *greenmasktransformers.RandomInt64Transformer
+}
+
+var ErrUnsupportedSizeError = errors.New("greenmask_integer: size must be 2, 4 or 8")
```
eminano, line 23: You can probably keep this unexported, since it's only used internally for now.

PR #155, opened 2025-02-21
Add gm float transformer

agedemenli:
Introduce `FloatTransformer` which uses `greenmasktransformers.RandomFloat64Transformer` under the hood.

### DIFF
--- a/pkg/transformers/greenmask/greenmask_float_transformer.go
+++ b/pkg/transformers/greenmask/greenmask_float_transformer.go
@@ -1,0 +1,82 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package greenmask
+
+import (
+	"fmt"
+	"math"
+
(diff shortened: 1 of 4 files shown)


pkg/transformers/greenmask/greenmask_float_transformer.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, transformers.ErrUnsupportedValueType
+	}
+	ret, err := ft.transformer.Transform(nil, toTransform)
+	if err != nil {
+		return nil, err
+	}
+	return float64(ret), nil
+}
+
+func getBytesForFloat(f float64) []byte {
```
eminano, line 70: It might be less verbose (and maybe a bit more readable) to use the encoding/binary library.
```
var buf [8]byte
binary.BigEndian.PutUint64(buf[:], math.Float64bits(f))
```
https://stackoverflow.com/a/43696189

eminano (approved):

Issue #128, opened 2025-02-05
Support snapshot only mode


eminano:
Currently pgstream supports continuous replication and snapshot+continuous replication modes. It should also support snapshot only mode, in case continuous replication is not possible due to limited access to the source postgres database, or in case only snapshots are required (no need to sync continuously).

The snapshot only mode shouldn't require keeping state on the source postgres database, in order to be as lightweight and unopinionated as possible.

PR #157, opened 2025-02-25
Add snapshot listener

eminano:
Add snapshot listener, an implementation of the wal listener that triggers a snapshot of a postgres database by using the generator adapter extracted in previous PRs (https://github.com/xataio/pgstream/pull/156).  Unlike the postgres wal listener, it doesn't start continuous replication.

A graceful shutdown has been added to the processor batch sender implementation to ensure that when the snapshot is completed, the service will wait for the processor to finish before exiting.

An option to disable triggers on the target postgres database is added to the configuration. This will use the `session_replication_role` to ensure foreign key and other triggers are not activated. This can help prevent snapshots from failing due to foreign key constraints or inconsistent data syncing by relying on the source postgres as the source of truth.

Readme is updated to add more examples and context around the snapshotting and the changes added in the context of #128.

PR has been split for ease of reviewing.


exekias (approved):
nice!

PR #158, opened 2025-02-27
Add UUID gm transformer

agedemenli:
Introduce `UUIDTransformer` which uses greenmask's `RandomUuidTransformer` under the hood

### DIFF
--- a/pkg/transformers/greenmask/greenmask_transformer.go
+++ b/pkg/transformers/greenmask/greenmask_transformer.go
@@ -6,34 +6,33 @@ import (
 	"time"
 
 	greenmaskgenerators "github.com/eminano/greenmask/pkg/generators"
-	greenmasktransformers "github.com/eminano/greenmask/pkg/generators/transformers"
 	"github.com/xataio/pgstream/pkg/transformers"
 )
 
-func setGenerator(t greenmasktransformers.Transformer, generatorType transformers.GeneratorType) error {
+func getGreenmaskGenerator(requiredByteLength int, generatorType transformers.GeneratorType) (greenmaskgenerators.Generator, error) {
 	// default to using random generator
-	if generatorType == "" {
+	if generatorType == "" || generatorType == transformers.Random {
 		generatorType = transformers.Random
 	}
 
 	var greenmaskGenerator greenmaskgenerators.Generator
 	switch generatorType {
 	case transformers.Random:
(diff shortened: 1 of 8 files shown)


pkg/transformers/greenmask/greenmask_transformer.go
```diff
@@ -6,11 +6,10 @@ import (
 	"time"
 
 	greenmaskgenerators "github.com/eminano/greenmask/pkg/generators"
-	greenmasktransformers "github.com/eminano/greenmask/pkg/generators/transformers"
 	"github.com/xataio/pgstream/pkg/transformers"
 )
 
-func setGenerator(t greenmasktransformers.Transformer, generatorType transformers.GeneratorType) error {
+func getGreenmaskGenerator(requiredByteLength int, generatorType transformers.GeneratorType) (greenmaskgenerators.Generator, error) {
```
eminano, line 12: Another way of handling this kind of situation to keep it flexible for potential new modifiers of the generator is to use the [options pattern](https://dev.to/kittipat1413/understanding-the-options-pattern-in-go-390c).

```
type generatorOption func(greenmaskgenerators.Generator) 

func setGenerator(t greenmasktransformers.Transformer, generatorType transformers.GeneratorType, opts ...generatorOption) error {
      var greenmaskGenerator greenmaskgenerators.Generator
      [...]
    
      for _, opt := range opts {
            opt(greenmaskGenerator)
      }  
      
    return t.SetGenerator(greenmaskGenerator)
}

func withHashReducer(byteLength int) generatorOption {
	return func(g greenmaskgenerators.Generator) {
		g = greenmaskgenerators.NewHashReducer(g, byteLength)
	}
}
```
You'll see it used often in constructors to add layers (decorators) that are optional or set internal fields that can be custom (like loggers). Here's an example:
https://github.com/xataio/pgstream/blob/main/pkg/wal/processor/postgres/postgres_batch_writer.go#L35
https://github.com/xataio/pgstream/blob/main/pkg/wal/processor/postgres/postgres_batch_writer.go#L45
https://github.com/xataio/pgstream/blob/main/pkg/wal/processor/postgres/postgres_batch_writer.go#L67-L69
https://github.com/xataio/pgstream/blob/main/pkg/wal/processor/postgres/postgres_batch_writer.go#L93-L105

eminano (approved):

PR #159, opened 2025-02-28
Add gm boolean transformer

agedemenli:
Introduce 1FloatTransformer` which uses Greenmask's `RandomFloat64Transformer` under the hood.


eminano (approved):
🚀

PR #160, opened 2025-03-03
Add choice gm transformer

agedemenli:
Introduce `ChoiceTransformer` which uses greenmask's `RandomChoiceTransformer` under the hood.

### DIFF
--- a/pkg/transformers/greenmask/greenmask_choice_transformer.go
+++ b/pkg/transformers/greenmask/greenmask_choice_transformer.go
@@ -1,0 +1,55 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package greenmask
+
+import (
+	"errors"
+	"fmt"
+
+	greenmasktransformers "github.com/eminano/greenmask/pkg/generators/transformers"
+	"github.com/eminano/greenmask/pkg/toolkit"
+	"github.com/xataio/pgstream/pkg/transformers"
+)
+
+type ChoiceTransformer struct {
+	transformer *greenmasktransformers.RandomChoiceTransformer
+}
+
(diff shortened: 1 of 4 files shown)


pkg/transformers/greenmask/greenmask_choice_transformer_test.go
```diff
@@ -0,0 +1,144 @@
...
+			require.NoError(t, err)
+			require.NotNil(t, transformer)
+			got, err := transformer.Transform(tt.input)
+			require.Equal(t, tt.wantErr, err)
+			if err != nil {
+				return
+			}
+			require.NotNil(t, got)
+			// if deterministic, check if we get the same result again
+			if tt.generatorType == transformers.Deterministic {
```
eminano, line 137: In this case, since it's a pool of choices, you can validate that the result is contained within the choices.

pkg/transformers/greenmask/greenmask_choice_transformer.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+type ChoiceTransformer struct {
+	transformer *greenmasktransformers.RandomChoiceTransformer
+}
+
+var errChoicesEmpty = errors.New("greenmask_choice: choices must not be empty")
+
+func NewChoiceTransformer(generatorType transformers.GeneratorType, params transformers.Parameters) (*ChoiceTransformer, error) {
+	choices := []*toolkit.RawValue{}
```
eminano, line 21: As discussed, the parameters will be of type `[]string` when coming from the yaml file, so it's at this stage that we should parse them into the `RawValue` type, which is greenmask specific.

eminano (approved):

PR #162, opened 2025-03-06
Add unix timestamp gm transformer

agedemenli:
Transform unix timestamps using greenmask's integer tranformer.

### DIFF
--- a/pkg/transformers/greenmask/greenmask_unix_timestamp_transformer.go
+++ b/pkg/transformers/greenmask/greenmask_unix_timestamp_transformer.go
@@ -1,0 +1,39 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package greenmask
+
+import (
+	"errors"
+	"fmt"
+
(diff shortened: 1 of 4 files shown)


pkg/transformers/greenmask/greenmask_unix_timestamp_transformer.go
```diff
@@ -0,0 +1,39 @@
...
+	maxValue, foundMax, err := transformers.FindParameter[int64](params, "max_value")
+	if err != nil {
+		return nil, fmt.Errorf("greenmask_unix_timestamp: max_value must be an int64: %w", err)
+	}
+
+	if !foundMin || !foundMax {
+		return nil, errMinMaxValueNotSpecified
+	}
+
+	transformer, err := NewIntegerTransformer(generatorType, transformers.Parameters{
```
eminano, line 29: Could we not use the params on input, add the size and leave the validation to the integer transformer? I think right now the min/max values are validated twice.

eminano (approved):

PR #164, opened 2025-03-11
Add ns email transformer

agedemenli:
Add `EmailTransformer` which uses neosync's email transformer

### DIFF
--- a/pkg/transformers/neosync/neosync_email_transformer.go
+++ b/pkg/transformers/neosync/neosync_email_transformer.go
@@ -45,7 +45,7 @@ func NewEmailTransformer(params transformers.Parameters) (*EmailTransformer, err
 
 	excludedDomains, err := findParameter[[]string](params, "excluded_domains")
 	if err != nil {
-		return nil, fmt.Errorf("neosync_email: excluded_domains must be type of any: %w", err)
+		return nil, fmt.Errorf("neosync_email: excluded_domains must be type of []string: %w", err)
 	}
 	var excludedDomainsAny any
 	if excludedDomains != nil {
--- a/pkg/transformers/neosync/neosync_email_transformer.go
+++ b/pkg/transformers/neosync/neosync_email_transformer.go
@@ -47,10 +47,6 @@ func NewEmailTransformer(params transformers.Parameters) (*EmailTransformer, err
 	if err != nil {
 		return nil, fmt.Errorf("neosync_email: excluded_domains must be type of []string: %w", err)
 	}
-	var excludedDomainsAny any
-	if excludedDomains != nil {
-		excludedDomainsAny = *excludedDomains
-	}
 
 	maxLength, err := findParameter[int](params, "max_length")
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/transformers/neosync/neosync_email_transformer.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("neosync_email: preserve_length must be a boolean: %w", err)
+	}
+
+	preserveDomain, err := findParameter[bool](params, "preserve_domain")
+	if err != nil {
+		return nil, fmt.Errorf("neosync_email: preserve_domain must be a boolean: %w", err)
+	}
+
+	excludedDomains, err := findParameter[any](params, "excluded_domains")
```
eminano, line 27: Might be good to add a comment here to say that neosync expects the `excluded_domains` as type `any`. Otherwise this should normally be a `[]string` and it can be misleading.

pkg/transformers/neosync/neosync_email_transformer.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("neosync_email: max_length must be an integer: %w", err)
+	}
+
+	seed, err := findParameter[int](params, "seed")
+	if err != nil {
+		return nil, fmt.Errorf("neosync_email: seed must be an integer: %w", err)
+	}
+
+	emailType, err := findParameter[string](params, "email_type")
```
eminano, line 42: Since this is an enum, we should add validation here, or the transform will fail if an invalid value is provided, when we could have caught it in the constructor. The same applies for the invalid email action param.

pkg/transformers/neosync/neosync_email_transformer_test.go
```diff
@@ -0,0 +1,119 @@
...
+	t.Parallel()
+	tests := []struct {
+		name   string
+		input  any
+		params transformers.Parameters
+
+		wantErr error
+	}{
+		{
+			name:    "ok - valid default parameters",
```
eminano, line 103: Maybe we could add tests that validate the output (by using the seed 0 we make it deterministic).

pkg/transformers/neosync/neosync_email_transformer.go
```diff
@@ -0,0 +1,89 @@
...
+	preserveDomain, err := findParameter[bool](params, "preserve_domain")
+	if err != nil {
+		return nil, fmt.Errorf("neosync_email: preserve_domain must be a boolean: %w", err)
+	}
+
+	excludedDomains, err := findParameter[[]string](params, "excluded_domains")
+	if err != nil {
+		return nil, fmt.Errorf("neosync_email: excluded_domains must be type of []string: %w", err)
+	}
+	var excludedDomainsAny any
```
eminano, line 50: Right now the value of `excludedDomainsAny` will never be a nil pointer, since you're passing a reference to a variable (a pointer to a nil value vs a nil pointer). You need to add a check so that the value passed to neosync is nil when the param is not found. It might be worth keeping it in a separate function (similar to the `toInt64Ptr` function), it makes it more readable, and it might come in handy later on.

```go 

func toAnyPtr(strArray *[]string) *any {
	if strArray == nil {
		return nil
	}

	var strArrayAny any
	strArrayAny = *strArray
	return &strArrayAny
}

```

eminano (approved):
Nice!

PR #174, opened 2025-03-14
[PostgreSQL processor] Handle errors gracefully

eminano:
This PR updates the query flushing behaviour of the postgres batch writer to ensure that if one of the queries within a batch fails with a non internal error (constraint violation, relation doesn't exist), the error is logged as dataloss, and the rest of the queries are retried and checkpointed, instead of stopping the processor at the first query error.

This aligns with the behaviour of all the other processors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ test:
 
 .PHONY: integration-test
 integration-test:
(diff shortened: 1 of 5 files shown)


agedemenli (approved):

agedemenli:
Tested it with the scenario that we discussed yesterday. It doesn't stop anymore <3

PR #195, opened 2025-03-24
Fix excluded_domains param for emails

agedemenli:


pkg/transformers/neosync/neosync_email_transformer.go
```diff
@@ -74,7 +74,7 @@ func NewEmailTransformer(params transformers.Parameters) (*EmailTransformer, err
 		return nil, errInvalidInvalidEmailAction
 	}
 
-	opts, err := neosynctransformers.NewTransformEmailOpts(preserveLength, preserveDomain, toAnyPtr(excludedDomains), toInt64Ptr(maxLength), toInt64Ptr(seed), emailType, invalidEmailAction)
+	opts, err := neosynctransformers.NewTransformEmailOpts(preserveLength, preserveDomain, toAnyPtr(&excludedDomains), toInt64Ptr(maxLength), toInt64Ptr(seed), emailType, invalidEmailAction)
```
eminano, line 77: Rather than passing a pointer to the excluded domains, now we should be changing the `toAnyPtr` to receive a `[]string` instead of `*[]string`. As it is right now we'll never pass a nil pointer, but a pointer to nil 😵 .

eminano (approved):
Just a small comment 🚀

PR #202, opened 2025-03-26
Add integration test for transformers

agedemenli:


pkg/stream/integration/pg_pg_integration_transformer_test.go
```diff
@@ -0,0 +1,184 @@
...
+					updatedAt:      time.Date(2023, 1, 1, 0, 0, 0, 0, time.UTC),
+					gender:         "male",
+				},
+			},
+		},
+	}
+
+	nextRowID := 1
+	for _, tc := range tests {
+		tc := tc
```
eminano, line 96: You shouldn't need this anymore. It was required before go 1.22.
https://go.dev/blog/loopvar-preview

pkg/stream/integration/pg_pg_integration_transformer_test.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	runStream(t, ctx, cfg)
+
+	testTable := "pg2pg_integration_transformer_test"
+	execQuery(t, ctx, fmt.Sprintf(createTableQuery, testTable))
+
+	targetConn, err := pglib.NewConn(ctx, targetPGURL)
```
eminano, line 65: I just realised the integration tests don't close the conns. While it's not a big deal, since the postgres instance is tear down after the test, it's probably good practice to just defer close them. If you don't mind adding it here and in the `execQuery` part please 🙏

eminano (approved):
Nice work! 🧪

Issue #203, opened 2025-03-26
Crash when using a transformer and the generated value doesn't fit in varchar(60)


tsg:
If the generated value doesn't fit the target character length (e.g. `varchar(60)`, it prints errors but ultimately crashes.

Happens with the northwinddb and this transformers config:

```
transformations:
  - schema: public
    table: customers
    column_transformers:
      contact_name:
        name: greenmask_firstname
      address:
        name: greenmask_string
        generator: deterministic
        min_length: 20
        max_length: 50
      phone:
        name: greenmask_string
        generator: deterministic
        symbols: "12345667890"
        min_length: 10
        max_length: 10
      fax:
        name: greenmask_string
        generator: deterministic
        symbols: "12345667890"
        min_length: 10
        max_length: 10
```

Here are the logs:

```
, phone, fax) VALUES($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11) "
2025-03-26T19:23:23.304059158Z DBG logger.go:33 > batching query args=["WELLI","Wellington Importadora","Elenora","Sales Manager","b8KwyUmiTebbmHCfSwnYtJEikhN4B1SlizvXVxmin8bc9LxzVnjUfccnIDgTxoZuKFjliO5C2TmjAwY","Resende","SP","08737-363","Brazil","urKC4hrR",null] commit_position=0/0 module=postgres_batch_writer sql="INSERT INTO \"public\".\"customers\"(customer_id, company_name, contact_name, contact_title, address, city, region, postal_code, country, phone, fax) VALUES($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11) "
2025-03-26T19:23:23.304125097Z DBG logger.go:33 > batching query args=["WHITC","White Clover Markets","Hilton","Owner","KYYk9Oiz8esfhGKAsGrZZNGgHslhJf4oxIPGZZxREnfLZZl0PjA9ftgiHLBtHs11","Seattle","WA","98128","USA","7nGwKmARP8Nl28IbUiXlRDGGttWU9KZFR1evyCEyw","b7Ypx95FobLPdKfZZLwtktWG7VijMrBsJk"] commit_position=0/0 module=postgres_batch_writer sql="INSERT INTO \"public\".\"customers\"(customer_id, company_name, contact_name, contact_title, address, city, region, postal_code, country, phone, fax) VALUES($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11) "
[...]

PR #213, opened 2025-03-27
Catch value too long error

agedemenli:
fixes: #203


eminano (approved):

internal/postgres/errors.go
```diff
@@ -61,6 +69,12 @@ func mapError(err error) error {
 				Details: pgErr.Message,
 			}
 		}
+		// Class 22 — Data Exception
+		if strings.HasPrefix(pgErr.Code, "22") {
+			return &ErrConstraintViolation{
```
eminano, line 74: This should be `ErrDataException`.

PR #217, opened 2025-03-28
Add masking support

agedemenli:


eminano (approved):
As discussed, this is probably ok for now. If we ever need more flexibility for masking other data types we can consider using a different library (like https://github.com/showa-93/go-mask).

eminano:
Ah, and don't forget to update the documentation as well! 📚 
https://github.com/xataio/pgstream/blob/main/docs/README.md#supported-transformers

PR #230, opened 2025-04-01
[Snapshot] `pg_dump` snapshot fixes

eminano:
This PR adds a few small bug fixes:
- Add quote identifiers to snapshot schema/tables
- Do not translate wildcard `*` tables for pgdump schema snapshots, since it prevents all resources in a given schema to be dumped by having a list of tables.


agedemenli (approved):
I see we do `schemaName+"."+tableName` in several places. Maybe we can add another util func for that, like:

```go
func QuoteQualifiedIdentifier(sch, tbl string) string {
	return pq.QuoteIdentifier(sc) + "." + pq.QuoteIdentifier(tb)
}
```

PR #237, opened 2025-04-02
Add random and deterministic generators

agedemenli:


pkg/transformers/generators/deterministic_bytes_generator.go
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package generators
+
+import (
+	greenmaskgenerators "github.com/eminano/greenmask/pkg/generators"
+)
+
+func NewDeterministicBytesGenerator(size int) (Generator, error) {
+	generator, err := greenmaskgenerators.GetHashBytesGen([]byte{}, size)
```
eminano, line 10: You can just return `greenmaskgenerators.GetHashBytesGen([]byte{}, size)` directly.

eminano (approved):

pkg/transformers/generators/deterministic_bytes_generator.go
agedemenli: why do you think we should ignore the error? It is possible to get an error, e.g in case size > 64

eminano: We're not ignoring it, we're returning it. The `greenmaskgenerators.GetHashBytesGen` function returns a generator and an error, so the extra handling should be redundant. Unless I'm missing something? 🤔

PR #240, opened 2025-04-03
[Transformers] Support dynamic parameters

eminano:
This PR updates the transformer interface to be able to support dynamic parameters when providing a transformation. The transformer interface now receives the value for the column to transform along with the map of all other row values so that dynamic transformations can be applied. For now, the dynamic parameters offer basic support to provide a column, but should be extended as we include templates.

Example of implementation can be found for the greenmask first name transformer, where the gender can be dynamically provided by a different column.

Example of yaml transformer rules, assuming a `test` table with columns `name` and `sex`:
```
transformations:
  - schema: public
    table: test
    column_transformers:
      name:
        name: greenmask_firstname
        dynamic_parameters:
          gender:
            column: sex
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/xataio/pgstream
 
 go 1.24.1
+
 require (
 	github.com/bytedance/sonic v1.13.2
 	github.com/cenkalti/backoff/v4 v4.3.0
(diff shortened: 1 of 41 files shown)


agedemenli (approved):
Looks great! Thanks 🚀

PR #249, opened 2025-04-04
Validate pg column types with transformer types

agedemenli:


pkg/transformers/transformer.go
```diff
@@ -9,6 +9,7 @@ import (
 
 type Transformer interface {
 	Transform(Value) (any, error)
+	Type() TransformerType
```
agedemenli, line 12: it makes sense to require all transformers have a type at the first glance but not sure if it's going to cost us any flexibility?

This is needed because I validated the transformer layer after it is created. An alternative might be doing the validation before the creation, just by checking the rules file.

Since we don't have the rules yet, in `stream_run.go`, (it is being parsed in `transformer.New()`), I implemented the validation on the transformer processor itself, without touching the creation phase of it. 
But it is also possible to move the parsing logic out of it, to `stream_run.go`, and do the validation there, directly on the rules. We can then call the creation logic after the validation.

eminano: Right now you're "leaking" the internal logic of the transformers in the `columnTransformerCompatibleWithPGType`. You're making a decision of what types the transformers are compatible with, completely outside of the scope of the transformers. 
Ideally, this is something that should be set and known by each transformer, and can be read from the outside. So I'd say rather than having a `Type` method, the `Transformer` interface should have a `GetCompatibleValues` method that returns non implementation specific types. This means, there should be no pg types in the transformer pkg. There should be transformer defined types (`String`, `Integer`, `Bool`, etc) that will then be mapped to postgres types by the transformer validator that you've implemented.  

The same applies to the current wal transformer, there should be no postgres logic there (it should work for any source as far as it knows). Instead, you should have an independent postgres transformer validator that validates the transformers supported types against the postgres types for the matching columns (by mapping the postgres types to the transformer types that are returned by the transformers `GetCompatibleValues` method. This can be a `Validate` interface function that is passed as an option to the wal processor by using the postgres implementation when using the postgres listener/snapshot.

pkg/wal/processor/transformer/wal_transformer_validator.go
```diff
@@ -0,0 +1,100 @@
...
+	"context"
+	"fmt"
+
+	"github.com/jackc/pgx/v5/pgtype"
+	pglib "github.com/xataio/pgstream/internal/postgres"
+	"github.com/xataio/pgstream/pkg/transformers"
+	"golang.org/x/exp/slices"
+)
+
+func GetTransformerValidationFunc(ctx context.Context, pgurl, snapshoturl string) func(*Transformer) error {
```
eminano, line 15: This should be done in the stream run, where we decide if we pass on the validation option to the transformer or not. You'd expose the `pgTransformerValidator` function and use it directly there when needed.

You have to think that the wal processor transformer layer doesn't know anything about pg urls or snapshot urls and they don't really mean anything in this context (it's leaking configuration details into the package).

pkg/wal/processor/transformer/wal_transformer.go
```diff
@@ -62,6 +64,16 @@ func WithLogger(l loglib.Logger) Option {
 	}
 }
 
+func WithValidator(validator func(*Transformer) error) Option {
```
eminano, line 67: The validator function doesn't need to have the full transformer as parameter, just the transformer map, so we can pass that only, since it will be used by the transformer internally, and it will have access to it even though it's unexported (you can export the `columnTransfomers` type for that).

`func(map[string]ColumnTransformers) error`

pkg/wal/processor/transformer/wal_transformer_validator.go
```diff
@@ -0,0 +1,100 @@
...
+			return nil
+		}
+		pgurl = snapshoturl
+	}
+	return pgTransformerValidator(ctx, pgurl)
+}
+
+func pgTransformerValidator(ctx context.Context, pgurl string) func(t *Transformer) error {
+	return func(t *Transformer) error {
+		conn, err := pglib.NewConn(ctx, pgurl)
```
eminano, line 28: This is ok since we're only calling this once. If we were performing queries frequently, we'd want to have a struct that would initialise a connection pool that can help reuse resources each time the validation is called. 
I think it's ok to leave it like this for now, but it's something to keep in mind.

pkg/wal/processor/transformer/wal_transformer_validator.go
```diff
@@ -0,0 +1,100 @@
...
+	if pgurl == "" {
+		if snapshoturl == "" {
+			return nil
+		}
+		pgurl = snapshoturl
+	}
+	return pgTransformerValidator(ctx, pgurl)
+}
+
+func pgTransformerValidator(ctx context.Context, pgurl string) func(t *Transformer) error {
```
eminano, line 26: You seem to use the validator signature often enough to benefit from having a typedef:
```
type ValidatorFn func(t *Transformer) error
```

pkg/stream/stream_run.go
```diff
@@ -221,11 +221,16 @@ func Run(ctx context.Context, logger loglib.Logger, config *Config, instrumentat
 
 	if config.Processor.Transformer != nil {
 		logger.Info("adding transformation layer to processor...")
-		transformer, err := transformer.New(config.Processor.Transformer, processor, transformer.WithLogger(logger))
+		validationFunc := transformer.GetTransformerValidationFunc(ctx, config.Listener.Postgres.Replication.PostgresURL, config.Listener.Snapshot.Generator.URL)
```
eminano, line 224: As mentioned in the other comments, it would be cleaner to have the checks in the `GetTransformerValidationFunc` here, by only adding the option if possible. It's also necessary to check for nil values on the config since only one listener is provided at a time.
```
opts := []transformer.Option{transformer.WithLogger(logger)}
pgURL := ""
switch {
  case config.Listener.Postgres != nil:
      pgURL = config.Listener.Postgres.Replication.PostgresURL
  case config.Listener.Snapshot != nil:
      pgURL =config.Listener.Snapshot.Generator.URL
}
if pgURL != "" {
  opts = append(opts, transformer.WithValidator(transformer.PostgresValidator(ctx, pgURL))
} 
```

pkg/wal/processor/transformer/wal_transformer_validator.go
```diff
@@ -0,0 +1,100 @@
...
+	return func(t *Transformer) error {
+		conn, err := pglib.NewConn(ctx, pgurl)
+		if err != nil {
+			return fmt.Errorf("creating postgres connection pool: %w", err)
```
eminano, line 30: This is a single connection rather than a pool, so this can be updated to remove the pool part (it can be misleading otherwise).

pkg/wal/processor/transformer/wal_transformer_validator.go
```diff
@@ -0,0 +1,100 @@
...
+				return fmt.Errorf("querying table rows: %w", err)
+			}
+			defer rows.Close()
+			fieldDescriptions := rows.FieldDescriptions()
+
+			// TODO: maybe error out if len(fieldDescriptions) != len(columnTransformers)
+			// if we start requiring a transformer for every column (noop transformers)
+
+			// map column names to column pg type OIDs, skip columns that don't have a transformer
+			mappedColumns := make(map[string]uint32, len(fieldDescriptions))
```
eminano, line 48: Do we need 2 loops here, or can it be done in one? 

```
        // check that all column transformers are compatible with corresponding column types
	for _, desc := range fieldDescriptions {
	        tr, found := columnTransformers[desc.Name]
		if !found {
			continue
		}
		if !pgTypeCompatibleWithTransformerType(tr.CompatibleTypes(),desc.DataTypeOID) {
			return fmt.Errorf("transformer specified for column %s does not support pg data type", colName)
		}
	}
```

pkg/transformers/transformer.go
```diff
@@ -47,6 +48,28 @@ const (
 	Masking                TransformerType = "masking"
 )
 
+type SupportedDataType string
+
+const (
+	ByteArray      SupportedDataType = "byte_array"
+	StringDataType SupportedDataType = "string"
```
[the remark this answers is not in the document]
agedemenli: I'd prefer the same but needed to add it because we already have `String` in that file, as transformer type:
`String                 TransformerType = "string"`

So this is just to differentiate them. Is there any better way to do it?

pkg/wal/processor/transformer/wal_transformer_validator.go
eminano, line 26: Oh and we need to add tests for this, there's currently none (the coverage 💀 💀 💀  gave it away 😅 ).

pkg/transformers/transformer.go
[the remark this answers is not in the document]
eminano: Ah it's true! Usually the consts have some identifiable part, so the `DataType` is not a bad idea. I guess it should be added to all of them for consistency in that case 🙈

pkg/wal/processor/transformer/wal_transformer_validator.go
agedemenli: it could be done in a single loop but I also wanted to validate the rules do not have any non-existing column.

eminano: Maybe add a comment to be explicit about what this is validating against.

eminano (approved):
Very nice work! 🚢

PR #264, opened 2025-04-10
Validate transformer parameters

agedemenli:


pkg/transformers/builder/transformer_builder.go
```diff
@@ -3,46 +3,115 @@
...
+		}
 		return transformers.NewMaskingTransformer(cfg.Parameters)
 	default:
 		return nil, transformers.ErrUnsupportedTransformer
 	}
 }
+
+// validateParameters checks if all provided parameters are in the expected set
+func validateParameters(provided map[string]any, expected []string) error {
+	expectedMap := make(map[string]struct{})
```
eminano, line 105: Super minor, but you can initialise a map with a given length to avoid reallocations. In this case we know the size, so it's good practice to reserve it. 
```expectedMap := make(map[string]struct{}, len(expected))```

File: pkg/transformers/builder/transformer_builder.go [hunk 2]
```diff
@@ -3,46 +3,115 @@
...
 	"github.com/xataio/pgstream/pkg/transformers/neosync"
 )
 
+// ErrUnknownParameter is returned when unknown parameter is provided
+var ErrUnknownParameter = fmt.Errorf("unknown parameter provided to transformer")
+
 func New(cfg *transformers.Config) (transformers.Transformer, error) {
 	switch cfg.Name {
 	case transformers.GreenmaskString:
+		if err := validateParameters(cfg.Parameters, greenmask.StringTransformerParams); err != nil {
```
eminano, line 19: Would it make sense to have the validation done within the transformer constructor, rather than exposing all the transformer parameters, which would otherwise be internal?
The function can be defined in the transformers package (like the other transformer parameter util functions), but be called within the constructor, which is the one that should make sure the input parameters are compatible with the implementation. 

Otherwise, imagine a situation where instead of using this builder, the constructor is called directly, and the validation doesn't happen.

eminano (approved):

PR #265, opened 2025-04-10
Add option to repeat snapshots when using snapshot recorder

eminano:
Add option to support repeatable snapshots when using the snapshot request recorder. Currently the behaviour is to filter out existing successfully completed snapshots to avoid performing them more than once, to fit the replication run with initial snapshot (usually it should only need to be performed once). However, for the snapshot mode only, we want to support having repeatable snapshots while keeping track of the requests and the status of all the runs.


agedemenli (approved):
Just one thing minor that caught my eye.

PR #283, opened 2025-04-15
Add NoopTransformer

agedemenli:
Noop transformers are expected to be used for security purposes only. We will be validating that all columns have assigned a transformer if the strict validation mode is enabled. In that case noop transformers can be used for columns that does not need to be actually transformed.

Noop transformers will be added to the map initially, for the transformer processor. But they will be deleted after the validation phase, as we don't want to be calling their `Transform` method every single time.

fixes: #259

### DIFF
--- a/pkg/transformers/builder/transformer_builder.go
+++ b/pkg/transformers/builder/transformer_builder.go
@@ -44,6 +44,8 @@ func New(cfg *transformers.Config) (transformers.Transformer, error) {
 		return transformers.NewPhoneNumberTransformer(cfg.Parameters)
 	case transformers.Masking:
 		return transformers.NewMaskingTransformer(cfg.Parameters)
+	case transformers.Noop:
+		return transformers.NewNoopTransformer()
 	default:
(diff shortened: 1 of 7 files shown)


pkg/transformers/builder/transformer_builder.go
```diff
@@ -44,6 +44,8 @@ func New(cfg *transformers.Config) (transformers.Transformer, error) {
 		return transformers.NewPhoneNumberTransformer(cfg.Parameters)
 	case transformers.Masking:
 		return transformers.NewMaskingTransformer(cfg.Parameters)
+	case transformers.Noop:
```
eminano, line 47: What if we didn't need a noop implementation at all? Could we keep the "noop" transformer as a configuration concept, that doesn't translate into the transformer implementation but it's caught early by the validation?

Maybe the validation step can receive the config as well as the transformer map and can validate both? Or we could potentially unify the parsing of the rules into the transformers map and the validation in one step?

Issue #260, opened 2025-04-09
[Transformers] All columns must have a transformer rule when PII mode enabled


eminano:
This can be a noop, or a transformer implementation, but all columns need to be part of the transformation rules to ensure there's no unintended data leaks.

This will only apply for snapshot mode for now, and the tables checked will be the ones configured to be snapshotted.

For replication mode to be supported, we need to first complete #112 so that we don't require transformers for all database tables.

PR #288, opened 2025-04-17
Add noop transformers support and Strict validation mode

agedemenli:
Single PR for both adding noop transformer support and strict validation mode.
Parsing the config, if no validation mode is not provided, the default will be `relaxed` mode. If validation mode is set to `strict` we'll be validating that all columns for that particular table are present in the transformer config, by comparing them to the column list we get from postgres.

fixes: #259 
fixes: #260


eminano (approved):
:chef-kiss: ✨

PR #291, opened 2025-04-23
[CLI] Split run and snapshot commands

eminano:
This PR updates the CLI to separate the run and the snapshot commands. They behave differently, with the `snapshot` performing a one off snapshot of the source into the configured target, while the `run` starts replication from the source into the target, with the option to perform an initial snapshot.


agedemenli (approved):
are we going to update the docs in a separate pr?

PR #315, opened 2025-04-30
Improve errors

### LINKED ISSUE #227: [Replication] Improve error log when replication slot doesn't exist
Currently we get an error `no rows` when starting pgstream without a replication slot created. Instead this should have a proper description so that it's easy to identify.

eminano:
This PR updates the errors for unmarshaling yaml configuration and no replication slot so that they're easier to understand. 

Closes #227

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -962,7 +962,7 @@ transformations:
   table_transformers: # List of table transformations
     - schema: <schema_name> # Name of the table schema
       table: <table_name> # Name of the table
-      validation_mode: <validation_mode> # To be used when the global validation_mode is set to `table_level`. Can be one of strict or relaxed. Defaults to relaxed if not provided
+      validation_mode: <validation_mode> # To be used when the global validation_mode is set to `table_level`. Can be one of strict or relaxed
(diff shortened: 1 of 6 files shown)


docs/README.md
```diff
@@ -958,14 +958,15 @@ The rules for the transformers are defined in a dedicated yaml file, with the fo
...
-      table: <table_name>
-      column_transformers:
-        <column_name>:
-          name: <transformer_name>
-          parameters:
+  validation_mode: <validation_mode> # Validation mode for the transformation rules. Can be one of strict, relaxed or table_level. Defaults to relaxed if not provided
+  table_transformers: # List of table transformations
+    - schema: <schema_name> # Name of the table schema
+      table: <table_name> # Name of the table
+      validation_mode: <validation_mode> # To be used when the global validation_mode is set to `table_level`. Can be one of strict or relaxed. Defaults to relaxed if not provided
```
agedemenli, line 965: It actually doesn't default to relaxed. If the global one is set to `table_level`, we expect every column to have a validation mode specified, error out otherwise.

agedemenli (approved):

Issue #311, opened 2025-04-29
Escape hatch transform


divyenduz:
If we don't support a type that needs to be anonymized. There should be a way to not return that column to onboard a user successfully. Example, let's say a user has heart rates in a GeoJSON activity stored in JSONB format. 

Escape hatches like
- masking: don't work for all types and doesn't adhere to uniqueness constraints
- uuid: doesn't work for all types (and therefore doesn't work for jsonb)

Maybe an escape hatch can be to hardcode a response, something like,

```
transformations:
  validation_mode: relaxed
  table_transformers:
    - schema: public
      table: Project
      column_transformers:
        heart_rates:
          name: literal or hardcode (I like literal more)
          parameters:
            value: '{}'
```

This will be a escape hatch that would work on all the types (though won't be useful when uniqueness constraint is present). This is just a suggestion, maybe there are other ways around it.

PR #318, opened 2025-04-30
Add literal string transformer

agedemenli:
Adding `LiteralStringTransformer` which transforms anything into given string. This is expected to be used as an escape hatch, mostly for non supported types.

fixes: #311

### DIFF
--- a/pkg/transformers/literal_string_transformer.go
+++ b/pkg/transformers/literal_string_transformer.go
@@ -1,0 +1,47 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package transformers
+
+import (
+	"errors"
+	"fmt"
+)
+
+type LiteralStringTransformer struct {
+	literal string
+}
+
+var (
+	literalStringTransformerParams = []string{"literal"}
+	errLiteralStringCannotBeEmpty  = errors.New("literal_string_transformer: literal parameter cannot be empty")
+)
+
+func NewLiteralStringTransformer(params Parameters) (*LiteralStringTransformer, error) {
+	if err := ValidateParameters(params, literalStringTransformerParams); err != nil {
+		return nil, err
+	}
+
+	literal, _, err := FindParameter[string](params, "literal")
+	if err != nil {
+		return nil, fmt.Errorf("literal_string_transformer: literal must be a string: %w", err)
+	}
+	if literal == "" {
+		return nil, errLiteralStringCannotBeEmpty
+	}
+
+	return &LiteralStringTransformer{
+		literal: literal,
+	}, nil
+}
+
+func (lst *LiteralStringTransformer) Transform(value Value) (any, error) {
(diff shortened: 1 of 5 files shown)


pkg/wal/processor/transformer/wal_transformer_validator.go
```diff
@@ -125,6 +125,8 @@ func pgTypeCompatibleWithTransformerType(compatibleTypes []transformers.Supporte
 		return slices.Contains(compatibleTypes, transformers.DateDataType)
 	case pgtype.TimestampOID, pgtype.TimestamptzOID:
 		return slices.Contains(compatibleTypes, transformers.DatetimeDataType)
+	case pgtype.JSONBOID:
+		return slices.Contains(compatibleTypes, transformers.JSONDataType)
```
agedemenli, line 129: I wonder what other types we might want this to be used with. Theoretically, just like jsonb, I'd expect it to work well with any pg type that can be represented as a string.

pkg/transformers/literal_string_transformer.go
```diff
@@ -0,0 +1,47 @@
...
+	if err := ValidateParameters(params, literalStringTransformerParams); err != nil {
+		return nil, err
+	}
+
+	literal, _, err := FindParameter[string](params, "literal")
+	if err != nil {
+		return nil, fmt.Errorf("literal_string_transformer: literal must be a string: %w", err)
+	}
+	if literal == "" {
+		return nil, errLiteralStringCannotBeEmpty
```
agedemenli, line 29: maybe we should allow empty strings, but keep disallowing not providing a string literal?

pkg/wal/processor/transformer/wal_transformer_validator.go
```diff
@@ -125,6 +125,8 @@ func pgTypeCompatibleWithTransformerType(compatibleTypes []transformers.Supporte
 		return slices.Contains(compatibleTypes, transformers.DateDataType)
 	case pgtype.TimestampOID, pgtype.TimestamptzOID:
 		return slices.Contains(compatibleTypes, transformers.DatetimeDataType)
+	case pgtype.JSONBOID:
```
eminano, line 128: We should add json type as well.

pkg/wal/processor/transformer/wal_transformer_validator.go
eminano: Yeah, this new transformer will only be compatible with strings, json and byte arrays, shouldn't we cast a much wider net adding all types that can be represented as strings? Arrays come to mind as well.

pkg/transformers/literal_string_transformer.go
eminano: Hmm, yeah, not sure if this should be an error, an empty value could be the desired behaviour.

PR #322, opened 2025-05-06
Update yaml examples

naveen246:


build/docker/docker-compose.yml
```diff
@@ -28,7 +29,7 @@ services:
     profiles: ["pg2pg"]
 
   kafka:
-    image: "bitnami/kafka:latest"
+    image: "bitnami/kafka:3.9.0"
```
naveen246, line 32: When I tried to run `pgstream run -c kafka2pg_tutorial.env --log-level trace`
I got the error - `Failed to join group pgstream-postgres-consumer-group: EOF`

The reason for the error seems to be - https://github.com/segmentio/kafka-go/issues/1360
segmentio/kafka-go does not seem to work with kafka 4.0 and using kafka 3.9.0 is working

eminano: We've tried this on our end in different computers and we didn't encounter this problem 🤔. We can't really explain why this happens in your case, maybe there's some difference in the setup somewhere? 

In any case, we'd rather keep the version as up to date as possible if there's no reproducible issue with it, if you don't mind removing the change 🙏

cmd/config/config_yaml.go
```diff
@@ -131,10 +131,11 @@ type ConstantBackoffConfig struct {
 }
 
 type PostgresTargetConfig struct {
-	URL              string       `mapstructure:"url" yaml:"url"`
-	Batch            *BatchConfig `mapstructure:"batch" yaml:"batch"`
-	DisableTriggers  bool         `mapstructure:"disable_triggers" yaml:"disable_triggers"`
-	OnConflictAction string       `mapstructure:"on_conflict_action" yaml:"on_conflict_action"`
+	URL               string       `mapstructure:"url" yaml:"url"`
+	Batch             *BatchConfig `mapstructure:"batch" yaml:"batch"`
+	SchemaLogStoreURL string       `mapstructure:"schema_log_store_url" yaml:"schema_log_store_url"`
```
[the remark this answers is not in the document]
eminano: Good catch! We missed it on the yaml, but you're right, it's supported with env variables and in the internal config already 😅

eminano (commented):
Thanks for updating the yaml examples and catching the missing config parameter @naveen246 🙌 !

If you wouldn't mind removing the change to the `kafka` docker image version, would be happy to merge!

build/docker/docker-compose.yml
naveen246: Thanks for checking, reverted the change.

PR #328, opened 2025-05-09
Add neosync lastname transformer

agedemenli:
Adding neosync's last name transformer.

### DIFF
--- a/pkg/transformers/neosync/neosync_lastname_transformer.go
+++ b/pkg/transformers/neosync/neosync_lastname_transformer.go
@@ -1,0 +1,52 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package neosync
+
+import (
+	"fmt"
+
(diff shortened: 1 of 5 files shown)


pkg/transformers/neosync/neosync_lastname_transformer.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	return &LastNameTransformer{
+		transformer: New[string](neosynctransformers.NewTransformLastName(), opts),
+	}, nil
+}
+
+func (t *LastNameTransformer) CompatibleTypes() []transformers.SupportedDataType {
+	return []transformers.SupportedDataType{
+		transformers.StringDataType,
```
eminano, line 50: I assume also byte array?

eminano (approved):

pkg/transformers/neosync/neosync_lastname_transformer.go
agedemenli: neosync's lastname transformer do not accept byte arrays, only string

PR #339, opened 2025-05-13
Add template transformer

agedemenli:
This PR adds template transformer which gets a template during creation, then executes it when transforming data. This transformer also supports dynamic values using other column values.

Other than the functionalities supported by default by templates, this PR introduces two methods, namely `GetValue` and `GetDynamicValue`, allowing users to build up more complex templates using all the column values in a row.

Considering adding support for wide range of helper functions supported [here](https://github.com/eminano/greenmask/blob/main/pkg/toolkit/template_functions.go#L71-L199) in greenmask, in a followup PR.
Skipping updating the docs for now, will do it once we're done with templates.


eminano (approved):
The possibilities! 🚀 Love this 💟

PR #342, opened 2025-05-16
Add basic profiling support to CLI

eminano:
This PR adds basic support for profiling for run/snapshot CLI commands. A flag `--profile` is added, which will create cpu and memory profile files if enabled, as well as start an http server exposing a `/debug/pprof` endpoint. The files will only be created on `snapshot` mode, since run is a long running process.

For now the values for the filenames and the server address are hardcoded to `mem.prof`, `cpu.prof` and `localhost:6060`.

### DIFF
--- a/cli-definition.json
+++ b/cli-definition.json
@@ -29,7 +29,7 @@
       "flags": [
         {
           "name": "profile",
-          "description": "Whether to expose a /debug/pprof/ endpoint on localhost:6060",
+          "description": "Whether to expose a /debug/pprof endpoint on localhost:6060",
           "default": "false"
         },
         {
@@ -84,7 +84,7 @@
         },
         {
           "name": "profile",
-          "description": "Whether to produce CPU and memory profile files, as well as exposing a /debug/pprof/ endpoint on localhost:6060",
+          "description": "Whether to produce CPU and memory profile files, as well as exposing a /debug/pprof endpoint on localhost:6060",
           "default": "false"
         },
         {
--- a/cli-definition.json
+++ b/cli-definition.json
@@ -29,7 +29,7 @@
       "flags": [
         {
           "name": "profile",
-          "description": "Wether to expose a /debug/pprof/ endpoint on localhost:6060",
+          "description": "Whether to expose a /debug/pprof/ endpoint on localhost:6060",
           "default": "false"
         },
         {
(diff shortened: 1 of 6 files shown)


cmd/root_cmd.go
```diff
@@ -59,6 +60,7 @@ func Prepare() *cobra.Command {
 	snapshotCmd.Flags().String("target-url", "", "Target URL")
 	snapshotCmd.Flags().StringSlice("tables", nil, "List of tables to snapshot, in the format <schema>.<table>. If not specified, the schema `public` will be assumed. Wildcards are supported")
 	snapshotCmd.Flags().Bool("reset", false, "Wether to reset the target before snapshotting (only for postgres target)")
+	snapshotCmd.Flags().Bool("profile", false, "Wether to produce CPU and memory profile files, as well as exposing a /debug/pprof/ endpoint on localhost:6060")
```
agedemenli, line 63: it looks like a typo at the first glance but it's "wether" for other flags as well 🤔

cli-definition.json
```diff
@@ -27,6 +27,11 @@
       "use": "run",
       "example": "\n\tpgstream run --source postgres --source-url <source-postgres-url> --target postgres --target-url <target-postgres-url>\n\tpgstream run --source postgres --source-url <source-postgres-url> --target postgres --target-url <target-postgres-url> --snapshot-tables <schema.table> --reset\n\tpgstream run --source kafka --source-url <kafka-url> --target elasticsearch --target-url <elasticsearch-url>\n\tpgstream run --source postgres --source-url <postgres-url> --target kafka --target-url <kafka-url>\n\tpgstream run --config config.yaml --log-level info\n\tpgstream run --config config.env",
       "flags": [
+        {
+          "name": "profile",
+          "description": "Wether to expose a /debug/pprof/ endpoint on localhost:6060",
```
agedemenli, line 32: why is there a "/" at the end?

agedemenli (approved):

cmd/root_cmd.go
eminano: Ah I'll fix it 👍

cli-definition.json
eminano: It's the root for other sub-endpoints (for example, `debug/pprof/block`). I'll remove it though.

Issue #343, opened 2025-05-19
Using masking for UUID results in a cryptic error


divyenduz:
Using masking on a UUID data type results in the following error. This issue tracks improving this error message. 

```
ERR logger.go:45 > creating transformer layer error.message="transformer specified for column 'id' in table \"public\".\"User\" does not support pg data type with oid: 2950"
```

PR #345, opened 2025-05-20
Improve err msg for data type incompatibility in transformer validation

agedemenli:
This PR improves the error message that is thrown when an incompatibility between postgres data type and the transformer configured is detected, by adding names of both the transformer and postgres data type.
An example, turning:
`transformer specified for column 'id' in table \"public\".\"test\" does not support pg data type with oid: xxx`
into:
`transformer 'string' specified for column 'id' in table \"public\".\"test\" does not support pg data type: int8`

fixes: #343


pkg/wal/processor/transformer/wal_postgres_transformer_parser.go
```diff
@@ -78,7 +79,11 @@ func (v *PostgresTransformerParser) ParseAndValidate(rules []TableRules) (map[st
 
 			// validate that the transformer is compatible with the column type
 			if !pgTypeCompatibleWithTransformerType(transformer.CompatibleTypes(), datatype) {
-				return nil, fmt.Errorf("transformer specified for column '%s' in table %s does not support pg data type with oid: %d", colName, tableKey, datatype)
+				typeForOid, ok := pgtype_map.TypeForOID(datatype)
```
divyenduz, line 82: Since `TypeForOID` is not defined in this PR, a general question: what kind of permissions does `pgstream` need on source except logical replication to be able to read pg catalog? Is this documented somewhere?

divyenduz (approved):
LGTM 🚀

pkg/wal/processor/transformer/wal_postgres_transformer_parser.go
```diff
@@ -33,6 +33,7 @@ func NewPostgresTransformerParser(ctx context.Context, pgURL string, builder tra
 
 func (v *PostgresTransformerParser) ParseAndValidate(rules []TableRules) (map[string]ColumnTransformers, error) {
 	transformerMap := map[string]ColumnTransformers{}
+	pgtypeMap := pgtype.NewMap()
```
eminano, line 36: This should probably be a field of the `PostgresTransformerParser`, since it only needs to be initialised once. In our case, the `ParseAndValidate` is only called once, but that might not always be the case.

Worth noting as well that the custom types will not be covered by this, for those the oid will still be in the error message.

eminano (approved):

pkg/wal/processor/transformer/wal_postgres_transformer_parser.go
agedemenli: not sure if I understood you question correctly, but this method doesn't query postgres to get the type. It's just a library that has the mapping oid->type name

PR #346, opened 2025-05-20
Allow empty transformer list config

agedemenli:
removing the empty check for transformers list

### DIFF
--- a/cmd/config/config_yaml.go
+++ b/cmd/config/config_yaml.go
@@ -634,9 +634,13 @@ func (c YAMLConfig) parseFilterConfig() *filter.Config {
 }
 
 func (c TransformationsConfig) parseTransformationConfig() (*transformer.Config, error) {
-	if len(c.TransformerRules) == 0 {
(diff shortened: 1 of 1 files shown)


cmd/config/config_yaml.go
```diff
@@ -634,11 +633,6 @@ func (c YAMLConfig) parseFilterConfig() *filter.Config {
 }
 
 func (c TransformationsConfig) parseTransformationConfig() (*transformer.Config, error) {
-	if len(c.TransformerRules) == 0 {
```
eminano, line 637: If there are no transformation rules we can skip the transformer layer altogether, so we should return `nil,nil` which should cover both yaml and env configs (since this method is used by both).

eminano: We can check if the rules are provided (which can be mandatory), by checking if the c.TransformerRules is `nil`, and return the error we were returning until now. And then do a check if the length is 0, in which case we just skip the transformation layer altogether.

eminano (approved):

Issue #347, opened 2025-05-21
JSON structure for transforms


divyenduz:
Like we did a JSON for CLI structure https://github.com/xataio/pgstream/pull/250
Which is consume in TypeScript with full type safety ([internal link](https://github.com/xataio/frontend/blob/main/external/xata-cli/src/lib/pgstream/commands.ts))

We should expose the available transforms and their parameters too.

exekias:
since this is user input, would it make sense for this to be japnschema or similar? that would help editors to validate files

PR #350, opened 2025-05-22
Refactor all transformers to introduce definitions

agedemenli:
This PR adds `TransformerDefinition` struct to package `builder`, which has the info of compatible types, parameters and the create function for every transformer defined. Also by adding a map of all transformers to their definitions, helps both when building transformers and also creating a json definition file to be used with `xata clone`.
```
type TransformerDefinition struct {
	SupportedTypes []transformers.SupportedDataType
	Parameters     []transformers.TransformerParameter
	CreateFunc     func(*transformers.Config) (transformers.Transformer, error)
}
```

With that, instead of having a huge `switch` statement for matching transformer names with their build functions, we can directly call the `CreateFunc` when building a transformer. Also, since now we have the parameters info during creation, we'll be verifying parameters right away, by calling `ValidateParameters` only once, instead of calling it at the beginning of every `New.....Transformer` function.

This map gets all the information (types and params) directly from each transformer. Use the `CreateFunc` defined in the map, will prevent any accidental case that we add a new transformer but forget to add it's definiton. Because in that case we'll get a `unexpected transformer name` error.

Also adding a `TransformerParameter` struct to package `transformers`, which is defined in each transformer file, to be used in the `builder`.
```
type TransformerParameter struct {
	Name          string
	SupportedType string
	Default       any
	Dynamic       bool
	Required      bool
}
```

Finally, adding a script `build-transformers-definition.go` that creates the json definition file for all transformers, using the map previously mentioned.

fixes: #347 

edit: Looks like copilot summarized the PR better than I do :D
> This PR refactors all transformer implementations to use structured definitions for supported parameters and compatible types while integrating them into a centralized transformer builder.
* Refactors parameter definitions from string slices to structured TransformerParameter arrays.
* Introduces corresponding CompatibleTypes and updates each transformer’s New* function accordingly.
* Updates the transformer builder to use a map-based approach and enhances Makefile to run the transformer definitions builder.

### DIFF
--- a/pkg/transformers/builder/transformer_builder.go
+++ b/pkg/transformers/builder/transformer_builder.go
@@ -18,13 +18,6 @@ type TransformerBuilder struct {
 
 type Option func(b *TransformerBuilder)
 
-type TransformerDefinition struct {
-	SupportedTypes []transformers.SupportedDataType
-	Parameters     []transformers.Parameter
-	Definition     *transformers.Definition
-	BuildFn        func(*transformers.Config) (transformers.Transformer, error)
-}
-
 func NewTransformerBuilder(opts ...Option) *TransformerBuilder {
 	b := &TransformerBuilder{}
 	for _, opt := range opts {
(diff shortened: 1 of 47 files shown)


transformers-definition.json
```diff
@@ -0,0 +1,614 @@
...
+      "name": "greenmask_firstname",
+      "supported_types": [
+        "string",
+        "byte_array"
+      ],
+      "parameters": [
+        {
+          "name": "generator",
+          "supported_type": "string",
+          "default": "random",
```
divyenduz, line 84: Can it list all the possible values of `generator` in anther column, like for this one the possible values are 
`random`, `deterministic` but the JSON structure doesn't contain that information.

agedemenli: done!

divyenduz (approved):
LGTM, thank you!

pkg/transformers/builder/transformer_builder.go
```diff
@@ -18,6 +18,12 @@ type TransformerBuilder struct {
 
 type Option func(b *TransformerBuilder)
 
+type TransformerDefinition struct {
```
eminano, line 21: I think this is a mix of two things. One, the definition of a transformer, and another the build function. I think this might be better off by separating the two concerns. The `TransformerDefinition` should probably move to the transformers root package (instead of the builder), and not contain the build function. 

```
type Definition struct {
	SupportedTypes []SupportedDataType
	Parameters     []TransformerParameter
}
```

pkg/transformers/transformer.go
```diff
@@ -82,6 +82,15 @@ const (
 	columnDynamicParam = "column"
 )
 
+type TransformerParameter struct {
```
eminano, line 85: I think if we're introducing this new concept, we should rename the old one, otherwise we'll have `transformers.Parameter` and `transformers.TransformerParameter` types, which are ambiguous and confusing. 

I'd suggest using `Parameter` for this new one you're adding, and renaming the existing one to `ParameterValues`, since it aligns better with what it represents.

pkg/transformers/greenmask/greenmask_boolean_transformer.go
```diff
@@ -13,12 +13,24 @@ type BooleanTransformer struct {
 	transformer *greenmasktransformers.RandomBoolean
 }
 
-var booleanTransformerParams = []string{"generator"}
+var (
+	BooleanParams = []transformers.TransformerParameter{
```
eminano, line 17: Rather than exposing all the transformer internals, maybe we can add a function that returns the transformer definition while keeping all these unexported? That function can then be called when building the map.

```
func BooleanTransformerDefinition() *transformers.Definition {
	return &transformers.Definition{
		SupportedTypes: booleanCompatibleTypes,
		Parameters:     booleanParams,
	}
}
```

This way the implementation details of the transformer are owned by the transformers package and can't be modified outside of it. Think of exported variables as global variables. Anything that uses this package has access to it and can modify it, which could result in unexpected behaviour down the line.

pkg/transformers/builder/transformer_builder.go
```diff
@@ -32,55 +38,170 @@ func WithInstrumentation(i *otel.Instrumentation) Option {
 	}
 }
 
+var TransformersMap = map[transformers.TransformerType]TransformerDefinition{
```
eminano, line 41: You can then create a map that contains both the definition and the build function.

```
var TransformersMap = map[transformers.TransformerType]struct {
	Definition *transformers.Definition
	BuildFn    func(cfg *transformers.Config) (transformers.Transformer, error)
}{
	transformers.GreenmaskBoolean: {
		Definition: greenmask.BooleanTransformerDefinition(),
		BuildFn: func(cfg *transformers.Config) (transformers.Transformer, error) {
		},
	},
	[...]
}
```
Alternatively you could keep the map simple (containing only the definitions), and keep the switch within the `New` for the build. This means we keep the concerns of the two separate (one is used for the map definitions, and the other for building transformers), although that means two places where new transformers should be added.

```
var TransformersMap = map[transformers.TransformerType]*transformers.Definition{
	transformers.GreenmaskBoolean: greenmask.BooleanTransformerDefinition(),
	[...]
}
```

PR #363, opened 2025-05-27
[Snapshots] Support negative values for table page count

eminano:
This PR updates the code to support negative values from the table count query, since postgres will return -1 when the table has no pages (empty).

### DIFF
--- a/pkg/snapshot/generator/postgres/data/pg_snapshot_generator.go
+++ b/pkg/snapshot/generator/postgres/data/pg_snapshot_generator.go
@@ -165,6 +165,10 @@ func (sg *SnapshotGenerator) snapshotTable(ctx context.Context, snapshotID strin
 	if err != nil {
 		return err
 	}
+	// the table is empty
+	if tablePageCount < 0 {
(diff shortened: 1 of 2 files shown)


pkg/snapshot/generator/postgres/data/pg_snapshot_generator.go
```diff
@@ -165,6 +165,10 @@ func (sg *SnapshotGenerator) snapshotTable(ctx context.Context, snapshotID strin
 	if err != nil {
 		return err
 	}
+	// the table is empty
+	if tablePageCount < 0 {
```
agedemenli, line 169: I wonder what does it mean if the page count is zero? Is it ever possible?

agedemenli (approved):

pkg/snapshot/generator/postgres/data/pg_snapshot_generator.go
eminano: Yeah, it means 1 page (they start on 0 for some reason).

Issue #145, opened 2025-02-17
[Anonymisation] Add json transformer support


eminano:
Support transforming fields within json column types.

PR #389, opened 2025-06-04
Json transformer support

agedemenli:
Adding JSON transformer for json and jsonb date types of postgres. This transformer executes a list of given operations on the json data to be transformed.
All operations must be either `set` or `delete`. 
`set` operations support literal values as well as templates, making use of `sprig` and `greenmask`'s function sets. Also, like the template transformer, `.GetValue` and `.GetDynamicValue` functions are added. Unlike template transformer, here `.GetValue` refers to the value at given path, rather than the entire JSON object; whereas `.GetDynamicValue` is again used for referring to other columns.
`delete` operations simply delete the object at the given path.
`Transform` errors out if the given path does not exists and the parameter `error_not_exist` - which is false by default - is set to true.

Adding helper function `getOperationsParam` to map the transformer parameters to the fields of struct `Operation`.
Also adding `JSONValue` to be used only within JSON transformer. As mentioned before, this one's `.GetValue` function - unlike the regular `Value` struct - refers to a particular path in the json object.
Using [sjson](https://github.com/tidwall/sjson) library for executing the operations. Just like `greenmask`, pathing should follow the [synxtax rules of sjson](https://github.com/tidwall/sjson#path-syntax)

Also adding tests and updating the docs.

fixes: #145

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1108,6 +1108,125 @@ transformations:
 #### Other Transformers
 
  <details>
+  <summary>json</summary>
+
+**Description:** Transforms json data with set and delete operations
+
+| Supported PostgreSQL types |
+| -------------------------- |
+| json, jsonb                |
+
+| Parameter   | Type   | Default | Required |
+| ----------- | ------ | ------- | -------- |
+| operations  | array  | N/A     | Yes      |
+
+Parameter for each operation:
+| Parameter       | Type    | Default | Required | Values                         |
+| --------------- | ------- | ------- | -------- | ------------------------------ |
+| operation       | string  | N/A     | Yes      | set, delete                    |
+| path            | string  | N/A     | Yes      | *sjson syntax                  |
+| error_not_exist | boolean | false   | No       | true, false                    |
+| value           | string  | N\A     | **Yes    | Any valid JSON representation  |
+| value_template  | string  | N\A     | **Yes    | Any template with valid syntax |
+
+*Paths should follow [sjson syntax](https://github.com/tidwall/sjson#path-syntax)
+
+**Either `value` or `value_template` must be provided if the operation is `set`. If both are provided, `value_template` takes precedence.
+
+JSON transformer can be used for Postgres types json and jsonb. This transformer executes a list of given operations on the json data to be transformed.
+
+All operations must be either `set` or `delete`.
+`set` operations support literal values as well as templates, making use of `sprig` and `greenmask`'s function sets. See template transformer section for more details. Also, like the template transformer, `.GetValue` and `.GetDynamicValue` functions are supported. Unlike template transformer, here `.GetValue` refers to the value at given path, rather than the entire JSON object; whereas `.GetDynamicValue` is again used for referring to other columns.
+`delete` operations simply delete the object at the given path.
+Execution of an operation errors out if the given path does not exists and the parameter `error_not_exist` - which is false by default - is set to true.
+
+JSON transformer uses [sjson](https://github.com/tidwall/sjson) library for executing the operations. Operation paths should follow the [synxtax rules of sjson](https://github.com/tidwall/sjson#path-syntax)
+
+With the below config `pgstream` transforms the json values in the column `user_info_json` of the table `users` by:
+- First, traversing all the items in the array named `purchases`, and for each element, setting value to "-" for key "item".
+- Then, deleting the object named "country" under the top-level object "address".
+- Completely masking the "city" value under object "address", using `go-masker`'s default masking function supported by `pgstream`'s templating.
+- Finally, setting the user's lastname after fetching it from some other column named `lastname`, using dynamic values support. Assuming there's such column, having the lastname info for users.
+
+Example input-output is given below the config.
+
+**Example Configuration:**
+
+```yaml
+transformations:
+  table_transformers:
+    - schema: public
+      table: users
+      column_transformers:
+        user_info_json:
+          name: json
+          parameters:
+            operations:
+              - operation: set
+                path: "purchases.#.item"
+                value: "-"
+                error_not_exists: true
+              - operation: delete
+                path: "address.country"
+              - operation: set
+                path: "address.city"
+                value_template:  "\"{{ masking \"default\" .GetValue }}\""
+              - operation: set
+                path: "user.lastname"
+                value_template:  "\"{{ .GetDynamicValue \"lastname\" }}\""
+```
+
+For input JSON value,
+```json
+{
+  "user": {
+    "firstname": "john",
+    "lastname": "unknown"
+  },
+  "residency": {
+    "city": "some city",
+    "country": "some country"
+  },
+  "purchases": [
+    {
+      "item": "book",
+      "price": 10
+    },
+    {
+      "item": "pen",
+      "price": 2
+    }
+  ]
+}
+```
+
+the JSON transformer with above config produces output:
+```json
+{
+  "user": {
+    "firstname": "john",
+    "lastname": "doe"
+  },
+  "residency": {
+    "city": "*********"
+  },
+  "purchases": [
+    {
+      "item": "-",
+      "price": 10
+    },
+    {
+      "item": "-",
+      "price": 2
+    }
+  ]
+}
+```
+
+</details>
+
+
+ <details>
   <summary>literal_string</summary>
 
 **Description:** Transforms all values into the given constant value.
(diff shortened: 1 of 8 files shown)


docs/README.md
```diff
@@ -1107,6 +1107,125 @@ transformations:
...
+| --------------- | ------- | ------- | -------- | ------------------------------ |
+| operation       | string  | N/A     | Yes      | set, delete                    |
+| path            | string  | N/A     | Yes      | *sjson syntax                  |
+| error_not_exist | boolean | false   | No       | true, false                    |
+| value           | string  | N\A     | **Yes    | Any valid JSON representation  |
+| value_template  | string  | N\A     | **Yes    | Any template with valid syntax |
+
+*Paths should follow [sjson syntax](https://github.com/tidwall/sjson#path-syntax)
+
+**Either `value` or `value_template` must be provided if the operation is `set`. If both are provided, `value_template` takes the precedende.
```
eminano:
```suggestion
**Either `value` or `value_template` must be provided if the operation is `set`. If both are provided, `value_template` takes precedence.
```

pkg/transformers/json_transformer.go
```diff
@@ -0,0 +1,275 @@
...
+	errOperationNameMustBeProvided   = errors.New("operation name must be provided in the operation definition")
+	errValueOrTemplateMustBeProvided = errors.New("either value or value_template must be provided in the 'set' operation definition")
+	errPathMustBeProvided            = errors.New("path must be provided in the operation definition")
+	jsonCompatibleTypes              = []SupportedDataType{
+		JSONDataType,
+	}
+	jsonParams = []Parameter{
+		{
+			Name:          "operations",
+			SupportedType: "array",
```
eminano, line 36: Would this be array of strings/text?

pkg/transformers/json_transformer.go
```diff
@@ -0,0 +1,275 @@
...
+			Default:       nil,
+			Dynamic:       false,
+			Required:      true,
+		},
+	}
+)
+
+type Operation struct {
```
eminano: I think the Operation and JSONValue logic would benefit from being in a separate file (`json_transformer_operation.go`), to make the transformer file itself more readable, keeping the variables at the top also more scoped to what's in the file. 

Side question, do `Operation` and `JSONValue` need to be exported?

pkg/transformers/json_transformer.go
```diff
@@ -0,0 +1,275 @@
...
+	dynValue, found := jv.dynamicValues[key]
+	if !found {
+		return nil, fmt.Errorf("dynamic value '%s' not found", key)
+	}
+	return dynValue, nil
+}
+
+type JSONTransformer struct {
+	operations []*Operation
+	tctx       *JSONValue
```
eminano: What does `tctx` stand for?

pkg/transformers/json_transformer.go
```diff
@@ -0,0 +1,275 @@
...
+func (jv *JSONValue) GetValue() any {
+	if !jv.exists {
+		return nil
+	}
+	return jv.value
+}
+
+func (jv *JSONValue) GetDynamicValue(key string) (any, error) {
+	if jv.dynamicValues == nil {
+		return nil, fmt.Errorf("dynamic values are nil")
```
eminano: You can make this error a variable at the top, and use `errors.New` since there's no formatting involved.

pkg/transformers/json_transformer.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+func (o *Operation) apply(inp []byte, tctx *JSONValue, buf *bytes.Buffer) ([]byte, error) {
+	var res []byte
+	var err error
+
+	switch o.Operation {
+	case jsonSetOpName:
+		if o.tmpl != nil {
+			buf.Reset()
```
eminano: You're resetting the buffer twice, you can probably do it just once.
Would it make sense to rather than having the buffer be part of the `Transformer` struct, to just initialise it here, since it only applies when the operation is a set and has a template? It feels a bit like a premature optimisation.

pkg/transformers/json_transformer.go
```diff
@@ -0,0 +1,275 @@
...
+	}, nil
+}
+
+func (jt *JSONTransformer) Transform(_ context.Context, value Value) (any, error) {
+	data, err := json.Marshal(value.TransformValue)
+	if err != nil {
+		return nil, fmt.Errorf("json_transformer: error marshalling value to JSON: %w", err)
+	}
+
+	jt.tctx.setDynamicValues(value.DynamicValues)
```
eminano: Some comments around here might be helpful, same for the variable names. `tctx` doesn't really give away meaning.

eminano: Not sure there's a need to have the `JSONValue` be part of the transformer struct, since it will be different on each transformation. It makes it seem like it's static, but its scope is the `Transform` function, and it's just being reused, which could lead to unexpected behaviour if any of the fields are not reset properly between transformations.

It might be cleaner to just create it here on each transform, with the dynamic values and the value on input, and then for each operation access the path if it exists. And have a separate struct that is the json sub element which is the one that will be potentially different between operations and needs to implement the `.GetValue` and `.GetDynamicValues` for the templates.

pkg/transformers/json_transformer.go
```diff
@@ -0,0 +1,275 @@
...
+
+	return &JSONTransformer{
+		operations: operations,
+		buf:        bytes.NewBuffer(nil),
+		tctx:       &JSONValue{},
+	}, nil
+}
+
+func (jt *JSONTransformer) Transform(_ context.Context, value Value) (any, error) {
+	data, err := json.Marshal(value.TransformValue)
```
eminano: Do you need to marshal/unmarshal at all? The value is likely to be already of []byte/string type for a json struct (already serialised). I'd say having a check for the type here will likely save us the expensive marshaling/unmarshaling in most cases.

On that note, try to use the internal [jsonlib](https://github.com/xataio/pgstream/blob/main/internal/json/json.go) when possible to use a faster implementation (it's a drop in replacement for the std json library).
It would not be a bad idea to extract some of the `sjson` and `gjson` functionality and add it to the jsonlib, to keep dependencies to a minimum. But this is a nice to have, it's ok as it is for now, since nothing else uses it.

pkg/transformers/json_transformer_test.go
```diff
@@ -0,0 +1,472 @@
...
+		dynamicValues map[string]any
+		params        ParameterValues
+
+		wantOutput         any
+		wantOutputContains string
+		wantErr            error
+	}{
+		{
+			name:  "ok - basic set and delete operations",
+			value: map[string]any{"greeting": "hello", "farewell": "goodbye"},
```
eminano, line 231: You should add a test with the []byte version as well, since that will likely be what we get on input. 
```
[]byte(`{"greeting":"hello", "farewell":"goodbye"}`)
```

pkg/transformers/json_transformer.go
agedemenli: I agree we should do it once, my bad, just deleted the one in `Transform` and kept the one in `apply`.
But I don't like the idea of initializing a buffer every single time executing a template. Actually I was considering doing the same thing in template transformer as well :D 
Does it sound like a bad idea to you?

pkg/transformers/json_transformer.go
agedemenli: For pg2pg scenario - tested on my local with a jsonb column - the value is of `map[string]any`.
Or, when the entire json object is an array, it is of `[]any`.

But if there's any case that we expect the value to be of `[]byte` or `string`, it could be worth to add the type check you mentioned. Or maybe just add it, just in case? 🤷‍♂️

eminano: Ah my bad, that makes sense, since we're unmarshaling the content to json, and it's an unknown struct so it'll be `map[string]any`. You can ignore that part of the comment then! 
We could add the check just in case, but probably for most of our cases it will need marshaling ;(

pkg/transformers/json_transformer.go
```diff
@@ -0,0 +1,197 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package transformers
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
```
eminano: You should still add the jsonlib for the marshaling/unmarshaling, it makes a considerable difference in performance, and this will be the hot path.

pkg/transformers/json_transformer.go
eminano: No, sounds ok. The main performance cost will be the json marshaling/unmarshaling here so not sure it will make a big difference. You can do something similar for the template transformer if you want 👍 
Another option could be using a buffer pool (example https://github.com/valyala/bytebufferpool). Similar goal, a bit easier to maintain.

PR #401, opened 2025-06-10
[Snapshot] Support multi-schema snapshots

eminano:
This PR updates the snapshot interface to support providing multiple schemas per snapshot. This will allow to perform the pg_dump of different schemas together to prevent missing dependencies, since it currently performs separate dumps that will fail if they are dependent on each other. 

This fixes https://github.com/xataio/xata-actions-demo/actions/runs/15486471181/job/43602182511.

The main changes are the switch from handling a single schema/tables to handling multiple. Most of the logic remains the same. Main changes:

- The wal snapshot adapter logic is moved to the data snapshot generator, since it's the one that will now parallelise the schema processing
- The pgdump/pgrestore schema snapshot generator is updated to provide the options for all schemas and their tables
- The snapshot requests remain at the schema level to maintain readability
- The snapshot errors are updated to register per schema errors

It's a pretty large PR, since the `Snapshot` type has been modified, so the commits are split to update each of the different dependent implementations separately.

### DIFF
--- a/pkg/snapshot/generator/postgres/data/pg_snapshot_generator_test.go
+++ b/pkg/snapshot/generator/postgres/data/pg_snapshot_generator_test.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sort"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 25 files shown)


pkg/snapshot/generator/postgres/data/pg_snapshot_generator_test.go
```diff
@@ -908,13 +909,15 @@ func TestSnapshotGenerator_CreateSnapshot(t *testing.T) {
...
+// 	if errors.As(err, &snapshotErrs) {
+// 		for _, schemaErrs := range snapshotErrs {
+// 			sort.Slice(schemaErrs.Tables, func(i, j int) bool {
+// 				return snapshotErrs.Tables[i].Table < snapshotErrs.Tables[j].Table
+// 			})
+// 		}
+// 		return snapshotErrs
+// 	}
+// 	return err
+// }
```
agedemenli: do we still want to keep this?

agedemenli (approved):
thank you!

PR #402, opened 2025-06-11
Handle fullname transformer edge cases

agedemenli:
Adding transformer parameters to `FullNameTransformer` to be used when handling edge cases.
Implementing cover scenarios instead of catching the errors because we know exactly when those errors happen beforehand.

With this PR, we'll be requiring `max_length` to be at least 3. When it's less than 5, we'll be generating random strings separated by a space, like "A Ab", as full names. Using the same randomizer with neosync transformer to ensure determinism.

When `preserve_length` is set to true and the input does not contain a last name (or when there's a single character last name) neosync's transformer produces buggy/incorrect output. See: https://github.com/nucleuscloud/neosync/issues/3536
With this PR, we'll be truncating the unnecessary part of the value.

Adding test cases to cover these edge cases.
Also updating the docs to mention `max_length` must be greater than 2, and `preserve_length` takes precedence when both are provided.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1057,6 +1057,8 @@ transformations:
 | max_length      | int  | 100     | No       |
 | seed            | int  | Rand    | No       |
 
+max_length must be greater than 2. If preserve_length is set to true, generated value can might be longer than max_length, depending on the input length.
+
 **Example Configuration:**
 
 ```yaml


pkg/transformers/neosync/neosync_fullname_transformer.go
```diff
@@ -40,34 +47,87 @@ var (
...
 	}
 
-	seed, err := findParameter[int](params, "seed")
+	maxLength := int(100)
+	if maxLengthParam != nil {
+		maxLength = *maxLengthParam
+	}
+
+	if maxLength < 2 {
+		return nil, errFullNameLengthMustBeGreaterThanTwo
```
eminano, line 74: If we're adding a space between first name and last name, and the min supported length for each is 1, shouldn't the max length be < 3?

pkg/transformers/neosync/neosync_fullname_transformer.go
```diff
@@ -40,34 +47,87 @@ var (
...
+		randomizer:     rng.New(seed),
+		maxLength:      maxLength,
+		preserveLength: preserveLength,
 	}, nil
 }
 
+func (t *FullNameTransformer) Transform(ctx context.Context, value transformers.Value) (any, error) {
+	var retVal string
+	maxLength := t.maxLength
+	if t.preserveLength {
```
eminano, line 103: Is there any way we can simplify this block now that the neosync fix for the truncate is coming? 
It feels like we should call the inner `Transform` and only change the returned value if it's not supported (like with the first name/last name transformers?

pkg/transformers/neosync/neosync_fullname_transformer.go
```diff
@@ -82,3 +142,26 @@ func FullNameTransformerDefinition() *transformers.Definition {
 		Parameters:     fullNameParams,
 	}
 }
+
+// generateRandomStringAsFullName generates a random full name string based on the specified length.
```
eminano, line 146: Another way of implementing this could be to use the neosync logic to generate a full name that is longer and then truncate the result to match the expected length (in this case 3 or 4).

agedemenli: In that case neosync errors out and we have nothing to truncate. We would simply get an error saying `unable to generate full name with length 3`
To get a longer fullname from neosync, we would need to create a new transformer with different opts. I think this implementation is simpler

pkg/transformers/neosync/neosync_fullname_transformer.go
agedemenli: yeah, my bad, fixing it

pkg/transformers/neosync/neosync_fullname_transformer.go
agedemenli: just deleted the truncate part. Keeping the random string generation part for no-preserve-length cases.

PR #405, opened 2025-06-12
[Snapshot] Data progress tracking

### LINKED ISSUE #286: Snapshot progress tracking
When a longer snapshot is under process, it would be cool if there is a way to provide progress status to the user. Currently, I am using `trace` log level BUT it is too noisy, specially when the data in rows in large (think `jsonb`). It would be cool if there is a mode that can do user readable progress, like

|Table|Progress|
|---|---|
|User | 4/20 |
|Post | 10/200 |

eminano:
This PR adds progress tracking to the data snapshots. The initial idea was to have a progress bar for each table being snapshotted. However, libraries supporting multiple progress bars ([multibar](https://github.com/sethgrid/multibar), [mpb](https://github.com/vbauerster/mpb) and [uiprogress](https://github.com/gosuri/uiprogress)) didn't show the progress properly, duplicating the result rather than updating the existing bar. 

Instead, this PR reports the progress for all tables in a schema in the same progress bar, using the [progressbar](https://github.com/schollz/progressbar) library.

The progress uses the total bytes for the snapshotted tables (using `pg_table_size` to avoid including indices which are snapshotted separately), and the average row bytes to calculate the progress made by each batch.

The logging of intermediate table snapshotting steps has been downgraded from `INFO` to `DEBUG` in favour of relying on the progress bar. 

Fixes #286

### DIFF
--- a/pkg/snapshot/generator/postgres/data/pg_snapshot_generator.go
+++ b/pkg/snapshot/generator/postgres/data/pg_snapshot_generator.go
@@ -6,11 +6,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 
 	"github.com/jackc/pgx/v5/pgconn"
 	pglib "github.com/xataio/pgstream/internal/postgres"
 	pglibinstrumentation "github.com/xataio/pgstream/internal/postgres/instrumentation"
+	"github.com/xataio/pgstream/internal/progress"
 	loglib "github.com/xataio/pgstream/pkg/log"
 	"github.com/xataio/pgstream/pkg/otel"
 	"github.com/xataio/pgstream/pkg/snapshot"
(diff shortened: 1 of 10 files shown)


pkg/snapshot/generator/postgres/data/pg_snapshot_generator.go
```diff
@@ -389,7 +459,33 @@ func (sg *SnapshotGenerator) getTablePageInfo(ctx context.Context, schemaName, t
 		return nil, err
 	}
 
-	return tablePageInfo, nil
+	return tableInfo, nil
+}
+
+const tablesBytesQuery = `SELECT SUM(pg_relation_size(c.oid)) FROM pg_class c JOIN pg_namespace n ON n.oid = c.relnamespace WHERE n.nspname = '%s' AND c.relname IN (%s) AND c.relkind = 'r';`
```
agedemenli: why do we use `pg_relation_size` here, to be used when adding progress bar; but use `pg_table_size` in `snapshotTable`?

eminano: Good catch, this should be `pg_table_size` 💯 ! I changed the previous PR and forgot to update it here 🙌

eminano: A super helpful [drawing](https://stackoverflow.com/a/70397779) to differentiate between all the options 😵:
image

PR #410, opened 2025-06-18
Add wildcard support for snapshot schemas

agedemenli:


pkg/snapshot/generator/postgres/tablefinder/pg_snapshot_table_finder.go
```diff
@@ -52,22 +57,34 @@ func NewSnapshotTableFinder(ctx context.Context, pgurl string, generator generat
...
+	if wildcardSchemaFound {
+		schemas, err := s.schemaDiscoveryFn(ctx, s.conn)
+		if err != nil {
+			return err
+		}
+		for _, schema := range schemas {
+			ss.SchemaTables[schema] = []string{wildcard}
+		}
+		delete(ss.SchemaTables, wildcard)
+	}
```
agedemenli: with this, we will not be specifically checking for `*.*`, but for `*.anything`
Since providing wildcard schema with a table name does no make sense, for now we're ignoring that case and treat it as `*.*`

Still not sure if we should expect the table part to be wildcard as well and maybe error out in case it's `*.table`?

pkg/snapshot/generator/postgres/tablefinder/pg_snapshot_table_finder.go
```diff
@@ -103,3 +120,27 @@ func discoverAllSchemaTables(ctx context.Context, conn pglib.Querier, schema str
 
 	return tableNames, nil
 }
+
+func discoverAllSchemas(ctx context.Context, conn pglib.Querier) ([]string, error) {
+	const query = "SELECT nspname FROM pg_catalog.pg_namespace WHERE nspname NOT IN ('pg_catalog', 'information_schema', 'pg_toast', 'pgstream')"
```
agedemenli, line 129: should we maybe not exclude `pgstream`?
This hardcoded query doesn't look good to me but I don't know how else we could do it 🤔

pkg/snapshot/generator/postgres/tablefinder/pg_snapshot_table_finder.go
exekias: I would rather error out at validation time if this is unexpected user input. They could expect that this filters for all tables named `table` on all schemas

pkg/snapshot/generator/postgres/tablefinder/pg_snapshot_table_finder.go
exekias: something to discuss, but would leave it out of the scope of this pr

exekias (approved):

PR #449, opened 2025-07-21
chore: fix inconsistent struct name in comment

stellrust:
fix inconsistent struct name in comment


eminano (approved):
Thank you for aligning the comments 🙌

Issue #443, opened 2025-07-17
[Snapshots] Add support for exclude tables filter


eminano:
Currently snapshots only support providing the list of tables that will be snapshotted. Add support for providing tables to exclude.

PR #452, opened 2025-07-23
Excluded tables support for snapshots

agedemenli:
Adding config parameter `excluded_tables` under snapshot. Provided tables will be removed from the snapshot schema tables map in the table finder layer, and will not be processed. Wildcards are not supported.

Also, changing the order of layers in `NewSnapshotGenerator` so that the finder layer (which now includes the exclusion as well) will be applied before schema processing. Otherwise we would end up creating the excluded table on the target, only the table itself though, not the data.

fixes: #443


cmd/config/config_yaml.go
```diff
@@ -461,6 +462,7 @@ func (c *YAMLConfig) parseSnapshotConfig() (*snapshotbuilder.SnapshotListenerCon
 			SchemaWorkers:   uint(snapshotConfig.Data.SchemaWorkers),
 			TableWorkers:    uint(snapshotConfig.Data.TableWorkers),
 			SnapshotWorkers: uint(snapshotConfig.SnapshotWorkers),
+			ExcludedTables:  snapshotConfig.ExcludedTables,
```
eminano: The excluded tables should be with the tables filter, in the adapter config. Then they would be passed along as part of the `snapshot.Snapshot` type, which would need to be updated to accommodate for the excluded tables filter, rather than being part of each of the generator objects. Also the excluded map logic wouldn't be duplicated, but would only be done once in the adapter like it's done for the [tables](https://github.com/xataio/pgstream/blob/main/pkg/wal/listener/snapshot/adapter/config.go#L15).
```
type Snapshot struct {
	SchemaTables map[string][]string
	SchemaExcludedTables map[string][]string
}
```

The generator objects should only have the fields that are common for any snapshot. Each snapshot could have a different schema/tables/excluded tables, so they should be parameters rather than struct fields.

pkg/snapshot/generator/postgres/schema/pgdumprestore/snapshot_pg_dump_restore_generator.go
```diff
@@ -346,6 +354,14 @@ func (s *SnapshotGenerator) pgdumpOptions(ctx context.Context, schemaTables map[
 		}
 	}
 
+	for schema, tables := range s.excludedTables {
```
eminano: Should add tests for this part.

cmd/config/config_yaml.go
eminano: Also, there are currently checks for when the tables are provided which will need to be updated to check for the excluded tables as well, since it's possible to only provide excluded tables now.
These are some examples, but I think there might be more places, worth double checking.
- https://github.com/xataio/pgstream/blob/main/cmd/config/config_env.go#L181
- https://github.com/xataio/pgstream/blob/main/cmd/run_cmd.go#L107

cmd/config/config_env.go
```diff
@@ -46,6 +46,7 @@ func init() {
 	viper.BindEnv("PGSTREAM_POSTGRES_SNAPSHOT_SCHEMA_WORKERS")
 	viper.BindEnv("PGSTREAM_POSTGRES_SNAPSHOT_TABLE_WORKERS")
 	viper.BindEnv("PGSTREAM_POSTGRES_SNAPSHOT_TABLES")
+	viper.BindEnv("PGSTREAM_POSTGRES_SNAPSHOT_EXCLUDED_TABLES")
```
eminano, line 49: Should add the new configuration option to the config tests too (both for yaml and env), and update the docs.

eminano (approved):
✨

PR #467, opened 2025-08-15
Introduce roles snapshotting

agedemenli:


pkg/snapshot/generator/postgres/schema/pgdumprestore/snapshot_pg_dump_restore_generator.go
```diff
@@ -177,8 +191,11 @@ func (s *SnapshotGenerator) CreateSnapshot(ctx context.Context, ss *snapshot.Sna
 	}
 
 	s.logger.Info("restoring schema indices and constraints", loglib.Fields{"schemaTables": ss.SchemaTables})
-	// apply the indices and constraints when the wrapped generator has finished
-	return s.restoreDump(ctx, dumpSchemas, append(dump.indicesAndConstraints, sequenceDump...))
+	// apply the indices, constraints and roles when the wrapped generator has finished
+	dumpsToRestore := dump.indicesAndConstraints
+	dumpsToRestore = append(dumpsToRestore, sequenceDump...)
+	dumpsToRestore = append(dumpsToRestore, rolesDump...)
```
eminano: Should the roles be created earlier, right after or even before the schema (since they are global objects)?

pkg/snapshot/generator/postgres/schema/pgdumprestore/snapshot_pg_dump_restore_generator.go
```diff
@@ -232,6 +249,21 @@ func (s *SnapshotGenerator) dumpSequenceValues(ctx context.Context, sequences []
 	return d, nil
 }
 
+func (s *SnapshotGenerator) dumpRoles(ctx context.Context) ([]byte, error) {
+	opts := &pglib.PGDumpAllOptions{
+		ConnectionString: s.sourceURL,
+		RolesOnly:        true,
```
eminano, line 252: The snapshot configuration allow to provide a role for the snapshot, which is used for the pg_dump command. It should be used in this one as well, for consistency. 
https://www.postgresql.org/docs/current/app-pg-dumpall.html
>--role=rolename
> Specifies a role name to be used to create the dump. This option causes pg_dumpall to issue a SET ROLE rolename command after connecting to the database. It is useful when the authenticated user (specified by -U) lacks privileges needed by pg_dumpall, but can switch to a role with the required rights. Some installations have a policy against logging in directly as a superuser, and use of this option allows dumps to be made without violating the policy.

internal/postgres/pg_dumpall.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"os/exec"
+)
+
+var pgDumpAllCmd = "pg_dumpall"
+
+type PGDumpAllOptions struct {
```
eminano, line 14: I noticed there's an option on `pg_dumpall` for the role password (`--no-role-passwords`). How is this currently handled, and will it work with the current settings if a role has a password?

pkg/snapshot/generator/postgres/schema/pgdumprestore/snapshot_pg_dump_restore_generator.go
```diff
@@ -545,6 +577,24 @@ func (s *SnapshotGenerator) parseDump(d []byte) *dump {
 	}
 }
 
+func removePostgresRole(rolesDump []byte) []byte {
+	// Match lines like CREATE ROLE postgres, ALTER ROLE postgres, DROP ROLE postgres (case-insensitive, allow extra whitespace)
+	roleRegexp := regexp.MustCompile(`(?i)^\s*(CREATE|ALTER|DROP)\s+ROLE\s+postgres\b`)
```
eminano: This regexp compile can be done at package level so it's done only once, otherwise it will be compiled on each call to the function.
Something to keep in mind, regular expressions are usually more expensive performance wise than string comparisons, so it's often preferred for hot path flows. In this case it will only be called once, so it's ok, but for non complex expressions it might be simpler to avoid them.

pkg/snapshot/generator/postgres/schema/pgdumprestore/snapshot_pg_dump_restore_generator.go
```diff
@@ -232,6 +249,21 @@ func (s *SnapshotGenerator) dumpSequenceValues(ctx context.Context, sequences []
...
+		RolesOnly:        true,
+	}
+
+	s.logger.Debug("dumping roles", loglib.Fields{"pg_dumpall_options": opts.ToArgs()})
+	d, err := s.pgDumpAllFn(ctx, *opts)
+	if err != nil {
+		s.logger.Error(err, "pg_dumpall for roles failed", loglib.Fields{"pgdumpallOptions": opts.ToArgs()})
+		return nil, fmt.Errorf("dumping roles: %w", err)
+	}
+	return removePostgresRole(d), nil
```
eminano: We are creating all roles in the source database (except for the `postgres` role). Ideally we would be dumping only those used by the selected schemas, is there any way of getting that information?
I can see people with a large number of unrelated roles in the source complaining about them appearing in the target when the snapshot is scoped to a given schema. It could even be considered a leak of internal source access data 👀 .

internal/postgres/pg_dumpall.go
agedemenli: Yes, it works.
We don't use that option at the moment and the dump looks like:
```
CREATE ROLE rolename;
ALTER ROLE rolename WITH NOSUPERUSER INHERIT NOCREATEROLE NOCREATEDB LOGIN NOREPLICATION NOBYPASSRLS PASSWORD 'SCRAM-SHA-256$4096:P2UpnfBMBE9UwoYKoYmFhA==$MxVUo.....................';
```

when the role is created like `CREATE ROLE rolename WITH PASSWORD 'mypassword' LOGIN;`

I can do the login on the target db using password. Authentication fails without the correct password.

pkg/snapshot/generator/postgres/schema/pgdumprestore/snapshot_pg_dump_restore_generator.go
agedemenli: Does "only those used by the selected schemas" mean that the roles that have permission on these schemas? Do you mean like excluding a role if it, say has the permission for only one schema which is not selected?

eminano: I think it might be a matter of checking the pgdump output, since that one contains the `SET ROLE` directives which will be used by the snapshot, and creating only those? This is what I think was causing all the errors and required to manually create the roles in the target for the snapshot to work.

agedemenli: I just dug deeper and noticed that the dump output does not only contain `SET ROLE` statements, but also other ones that depend on roles too, such as `GRANT ... ON SCHEMA ... TO rolename;` or
`ALTER TABLE ... OWNER TO rolename;`

I don't think it's a good idea to iterate through the dump and try to detect all these statements that depend on a role.
We can either keep it as is and snapshot all the roles; or ask user to provide names for the needed roles in a config or such, but that doesn't sound like a bright idea either.

eminano: I'd say it's valid to parse the output to identify the roles required for the snapshot. I wouldn't keep all source roles replicated to the target, and asking the user to provide the role names is not a solution when there are multiple roles and this needs to be done manually. 
The best way would be to have it be automatically detected, and since the roles are global, I'm not sure there's any other way of knowing which ones are needed other than parsing the schema dump.

agedemenli: ###  Part 1- Done
[This commit](https://github.com/xataio/pgstream/pull/467/commits/40ab831213d78b13393ed944e9516f014401cc78) adds the necessary logic to parse and catch (some of) the statements in the dump to find the names of needed roles, for below statements:
* `ALTER ... OWNER TO <rolename>;`
* `GRANT ..... TO <rolename>;`
* `REVOKE ..... FROM <rolename>;`
* `SET ROLE <rolename>;`

### Part 2- Ignored for now
I tested and verified that pg_dump or pg_dumpall does not generate below statements. So it's ok to ignore them.
* `DROP OWNED BY <rolename>;`
* `REASSIGN OWNED BY <rolename>;` 

### Part 3- In progress, simple
I haven't added the case for `COMMENT ON ROLE <rolename> IS ...` statements yet, but it should be straightforward.

### Part 4- In progress, tricky
The only tricky point at the moment is that `GRANT` statements can also be for assigning roles to roles. If the user has executed a command on the source, like `GRANT r1 to r2;`, it will show up in the output of `pg_dumpall --roles-only`. But if these roles have nothing to do with the schema to be snapshotted, they will not be mentioned in the actual dump, so we will not be able to detect them and add the `roles` map, so `pg_restore` will fail with error `role r1 does not exist`. The only solution that comes to my mind is to parse the roles dump as well to catch this case and do not skip the `CREATE ROLE` lines for these two.

### Part 5- Haven't started yet
I don't know if the below statements can be in the output of `pg_dump` or `pg_dumpall`, will have a look for them as well.
* `SET SESSION AUTHORIZATION role_name`
* `RESET SESSION AUTHORIZATION`
* `ALTER DEFAULT PRIVILEGES FOR ROLE role_name ...`

pkg/snapshot/generator/postgres/schema/pgdumprestore/snapshot_pg_dump_restore_generator.go
```diff
@@ -520,29 +553,111 @@ func (s *SnapshotGenerator) parseDump(d []byte) *dump {
 			strings.HasPrefix(line, "COMMENT ON TRIGGER"):
 			indicesAndConstraints.WriteString(line)
 			indicesAndConstraints.WriteString("\n\n")
+			continue // skip, do not add it to the filtered dump
```
eminano: It's not necessary to add `continue` to a switch. Only the case that matches first will be executed, unless there's a [`fallthrough`](https://go.dev/ref/spec#Expression_switches)(which is not the case here).

pkg/snapshot/generator/postgres/schema/pgdumprestore/snapshot_pg_dump_restore_generator.go
```diff
@@ -520,29 +553,111 @@ func (s *SnapshotGenerator) parseDump(d []byte) *dump {
...
+				roleName2 := getRoleNameAfterClause(secondPart, " TO ")
+				roleNames[roleName] = struct{}{}
+				roleNames[roleName2] = struct{}{}
+			} else if strings.Contains(line, " ON ") {
+				// GRANT <privileges> ON <object> TO <rolename>;
+				roleNames[getRoleNameAfterClause(line, " TO ")] = struct{}{}
+			}
+		case strings.HasPrefix(line, "REVOKE "):
+			roleNames[getRoleNameAfterClause(line, " FROM ")] = struct{}{}
+		case strings.HasPrefix(line, "SET ROLE ") || strings.HasPrefix(line, "SET SESSION ROLE ") || strings.HasPrefix(line, "SET LOCAL ROLE "):
```
eminano: In switch `case` you don't need to use OR operator, you can list the cases that will implement the same logic followed by a ",":
```
case strings.HasPrefix(line, "SET ROLE"), 
strings.HasPrefix(line, "SET SESSION ROLE "), 
strings.HasPrefix(line, "SET LOCAL ROLE):
```

pkg/snapshot/generator/postgres/schema/pgdumprestore/snapshot_pg_dump_restore_generator.go
```diff
@@ -520,29 +553,111 @@ func (s *SnapshotGenerator) parseDump(d []byte) *dump {
...
+				_, afterGrant, _ := strings.Cut(afterRole, " GRANT ")
+				_, afterOn, _ := strings.Cut(afterGrant, " ON ")
+				roleNames[getRoleNameAfterClause(afterOn, " TO ")] = struct{}{}
+			}
 		case strings.HasPrefix(line, "CREATE SEQUENCE"):
 			qualifiedName, err := pglib.NewQualifiedName(strings.TrimPrefix(line, "CREATE SEQUENCE "))
 			if err == nil {
 				sequenceNames = append(sequenceNames, qualifiedName.String())
 			}
-			fallthrough
```
eminano, line 533: I think you wanted to avoid adding the write string to the filteredDump instead of having a deafult. I think this change makes it a lot less readable, and prone to error, polluting the switch case with unusual language patterns (`continue`). I'd recommend just adding the write string for the line to whichever dump necessary.
An alternative for readability/keeping it dry is to have a function that encapsulates all the roles checks, and it is represented as a single case in this switch. 

```
case isRoleStatement():
   addRoleNames(line, roleNames)
   filteredDump.WriteString(line)
   filteredDump.WriteString("\n")
```

pkg/snapshot/generator/postgres/schema/pgdumprestore/snapshot_pg_dump_restore_generator.go
```diff
@@ -520,29 +553,111 @@ func (s *SnapshotGenerator) parseDump(d []byte) *dump {
...
-		default:
-			filteredDump.WriteString(line)
-			filteredDump.WriteString("\n")
 		}
+
+		filteredDump.WriteString(line)
+		filteredDump.WriteString("\n")
 	}
 
+	delete(roleNames, "postgres") // remove the postgres role from the list, since it is not needed and can cause issues
```
eminano, line 629: We could also just not add the role instead of removing it later.

eminano (approved):

PR #473, opened 2025-08-25
[Transformers] Add support for citext type to email transformer

### LINKED ISSUE #471: Transformer citext support
Using `xata clone`:

`Error: transformer 'neosync_email' specified for column does not support pg data type with oid: 16992`

My email columns are `citext`. I'm not sure how easy it would be for the validation to treat these columns as strings. I did some initial digging and found this: https://github.com/jackc/pgtype/issues/1

Alternately is there an option to skip column type validation?

eminano:
This PR updates the neosync email transformer to support citext postgres type. 
It also updates the validator handling to check against data type name when the OID is unknown to cover for extension/custom types.

Fixes #471

### DIFF
--- a/pkg/transformers/neosync/neosync_email_transformer.go
+++ b/pkg/transformers/neosync/neosync_email_transformer.go
@@ -86,6 +86,7 @@ var (
 	}
 	emailCompatibleTypes = []transformers.SupportedDataType{
 		transformers.StringDataType,
+		transformers.CitextDataType,
 	}
 )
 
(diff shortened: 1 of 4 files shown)


agedemenli (approved):
should we mention `citext` in the docs as a supported type for `neosync_email` transformer?

gulcin:
> should we mention `citext` in the docs as a supported type for `neosync_email` transformer?

Yes, that would be good, thanks @agedemenli !

Issue #468, opened 2025-08-19
`pgstream init` to be idempotent


divyenduz:
Reuse the replication slot (default name or provided), if it already exists.

PR #476, opened 2025-08-26
Command init to be idempotent in case replication slot exists

agedemenli:
Adds a check if the replication slot already exists, before creating the pgstream schema and the slot.

closes: #468

### DIFF
--- a/pkg/stream/stream_init.go
+++ b/pkg/stream/stream_init.go
@@ -25,7 +25,7 @@ const (
 var errMissingPostgresURL = errors.New("postgres URL is required")
 
 // Init initialises the pgstream state in the postgres database provided, along
-// with creating the relevant replication slot.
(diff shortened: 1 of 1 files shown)


pkg/stream/stream_init.go
```diff
@@ -37,6 +37,22 @@ func Init(ctx context.Context, pgURL, replicationSlotName string) error {
...
+			return err
+		}
+	}
+
+	// check if the replication slot already exists
+	exists, err := replicationSlotExists(ctx, conn, replicationSlotName)
+	if err != nil {
+		return fmt.Errorf("failed to check if replication slot exists: %w", err)
+	}
+	if exists {
```
eminano, line 67: It is possible the replication slot exists, but the schema or the migrations don't. This check should only apply to the creation of the replication slot, not the other steps (they are independent states).

Issue #475, opened 2025-08-26
Make roles snapshotting optional


agedemenli:
Add a new config param, something like `include_roles` which is set to `true` by default, maybe under `snapshot.schema.pgdump_pgrestore`?
Use `pgdump_all --roles-only` to snapshot roles only when this config is `true`.

This is to enable users to skip roles snapshotting in case they get blocked by it somehow.

PR #480, opened 2025-08-27
Add config param to optionally disable roles snapshotting

agedemenli:
Adds param `roles_snapshot_mode` to optionally disable roles snapshotting, or to exclude passwords from the dump. If it's set to 'disabled', roles will not be snapshotted. If it's set to 'no_passwords' roles will be snapshotted, but their passwords will not.

closes: #475

### DIFF
--- a/pkg/snapshot/generator/postgres/schema/pgdumprestore/snapshot_pg_dump_restore_generator_test.go
+++ b/pkg/snapshot/generator/postgres/schema/pgdumprestore/snapshot_pg_dump_restore_generator_test.go
@@ -493,6 +493,7 @@ func TestSnapshotGenerator_CreateSnapshot(t *testing.T) {
 					require.Equal(t, pglib.PGDumpAllOptions{
 						ConnectionString: "source-url",
 						RolesOnly:        true,
+						NoPasswords:      false,
 					}, po)
 					return rolesDumpOriginal, nil
 				default:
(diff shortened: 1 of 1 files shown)


cmd/config/config_yaml.go
```diff
@@ -512,6 +513,7 @@ func (c *YAMLConfig) parseSchemaSnapshotConfig() (snapshotbuilder.SchemaSnapshot
 			streamSchemaCfg.DumpRestore.IncludeGlobalDBObjects = schemaSnapshotCfg.PgDumpPgRestore.IncludeGlobalDBObjects
 			streamSchemaCfg.DumpRestore.CreateTargetDB = schemaSnapshotCfg.PgDumpPgRestore.CreateTargetDB
 			streamSchemaCfg.DumpRestore.Role = schemaSnapshotCfg.PgDumpPgRestore.Role
+			streamSchemaCfg.DumpRestore.RolesSnapshotMode = schemaSnapshotCfg.PgDumpPgRestore.RolesSnapshotMode
```
eminano: Maybe we can add a validation for the role snapshot mode, so that only valid values are accepted, rather than setting anything and applying the default silently.

pkg/snapshot/generator/postgres/schema/pgdumprestore/snapshot_pg_dump_restore_generator.go
```diff
@@ -86,6 +89,7 @@ func NewSnapshotGenerator(ctx context.Context, c *Config, opts ...Option) (*Snap
 		createTargetDB:         c.CreateTargetDB,
 		includeGlobalDBObjects: c.IncludeGlobalDBObjects,
 		role:                   c.Role,
+		rolesSnapshotMode:      c.RolesSnapshotMode,
```
eminano, line 92: Validation of the mode could also be done here.

pkg/snapshot/generator/postgres/schema/pgdumprestore/snapshot_pg_dump_restore_generator_test.go
```diff
@@ -455,6 +456,60 @@ func TestSnapshotGenerator_CreateSnapshot(t *testing.T) {
 
 			wantErr: nil,
 		},
+		{
+			name:              "ok - no roles dump",
```
eminano, line 460: Should add a test for no passwords mode to confirm we're passing the right options to the pgdump all.

eminano (approved):

cmd/config/config_yaml.go
```diff
@@ -513,8 +521,16 @@ func (c *YAMLConfig) parseSchemaSnapshotConfig() (snapshotbuilder.SchemaSnapshot
 			streamSchemaCfg.DumpRestore.IncludeGlobalDBObjects = schemaSnapshotCfg.PgDumpPgRestore.IncludeGlobalDBObjects
 			streamSchemaCfg.DumpRestore.CreateTargetDB = schemaSnapshotCfg.PgDumpPgRestore.CreateTargetDB
 			streamSchemaCfg.DumpRestore.Role = schemaSnapshotCfg.PgDumpPgRestore.Role
-			streamSchemaCfg.DumpRestore.RolesSnapshotMode = schemaSnapshotCfg.PgDumpPgRestore.RolesSnapshotMode
 			streamSchemaCfg.DumpRestore.DumpDebugFile = schemaSnapshotCfg.PgDumpPgRestore.DumpFile
+
+			switch schemaSnapshotCfg.PgDumpPgRestore.RolesSnapshotMode {
```
eminano, line 526: The config validation needs to apply to the env config as well (it's currently only for the yaml one).

Issue #469, opened 2025-08-19
`pgstream run --init` flag


divyenduz:
Add `--init` flag to `pgstream run` command that runs idempotent  init before starting the replication.

PR #481, opened 2025-08-27
Add --init option to run command

agedemenli:
Adding `--init` option to `run` command to run `init` before starting the stream.

closes: #469

### DIFF
--- a/cmd/root_cmd.go
+++ b/cmd/root_cmd.go
@@ -72,7 +72,7 @@ func Prepare() *cobra.Command {
 	runCmd.Flags().StringSlice("snapshot-tables", nil, "List of tables to snapshot if initial snapshot is required, in the format <schema>.<table>. If not specified, the schema `public` will be assumed. Wildcards are supported")
 	runCmd.Flags().Bool("reset", false, "Whether to reset the target before snapshotting (only for postgres target)")
 	runCmd.Flags().Bool("profile", false, "Whether to expose a /debug/pprof endpoint on localhost:6060")
-	runCmd.Flags().BoolVar(&initFlag, "init", false, "Whether to initialize the pgstream")
+	runCmd.Flags().BoolVar(&initFlag, "init", false, "Whether to initialize pgstream before starting replication")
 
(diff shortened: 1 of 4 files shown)


cmd/root_cmd.go
```diff
@@ -72,6 +72,7 @@ func Prepare() *cobra.Command {
 	runCmd.Flags().StringSlice("snapshot-tables", nil, "List of tables to snapshot if initial snapshot is required, in the format <schema>.<table>. If not specified, the schema `public` will be assumed. Wildcards are supported")
 	runCmd.Flags().Bool("reset", false, "Whether to reset the target before snapshotting (only for postgres target)")
 	runCmd.Flags().Bool("profile", false, "Whether to expose a /debug/pprof endpoint on localhost:6060")
+	runCmd.Flags().BoolVar(&initFlag, "init", false, "Whether to initialize the pgstream")
```
eminano:
```suggestion
	runCmd.Flags().BoolVar(&initFlag, "init", false, "Whether to initialize pgstream before starting replication")
```

pkg/stream/stream_run.go
```diff
@@ -27,11 +27,17 @@ import (
...
 
 // Run will run the configured pgstream processes. This call is blocking.
-func Run(ctx context.Context, logger loglib.Logger, config *Config, instrumentation *otel.Instrumentation) error {
+func Run(ctx context.Context, logger loglib.Logger, config *Config, init bool, instrumentation *otel.Instrumentation) error {
 	if err := config.IsValid(); err != nil {
 		return fmt.Errorf("incompatible configuration: %w", err)
 	}
 
+	if init {
+		if err := Init(context.Background(), config.SourcePostgresURL(), config.PostgresReplicationSlot()); err != nil {
```
eminano: Here you should pass the context on input.

eminano (approved):
Might be worth adding it to the readme usage as well, maybe under the prepare the database section?
https://github.com/xataio/pgstream?tab=readme-ov-file#prepare-the-database

Issue #482, opened 2025-08-28
[Snapshots] Roles should also be deleted when `clean_target_db` is enabled


eminano:
Currently the snapshot will create the relevant roles in the target by default. However, unlike with the rest of the schema/data, it will not clean the roles on each new snapshot call, showing dump ignored errors `role <role> already exists`. This should be avoided by using the `--clean` flag in the `pg_dumpall` command used to snapshot the roles, so that they are deleted on each snapshot call, similar to what's done for [pg_dump](https://github.com/xataio/pgstream/blob/main/internal/postgres/pg_dump.go#L36). 

https://www.postgresql.org/docs/current/app-pg-dumpall.html

PR #485, opened 2025-08-29
Clean db before dumping roles, if clean_target_db is set

agedemenli:
Add option `--clean` to `pg_dumpall` if the config param `clean_target_db` is set. Otherwise we get error `role already exists`.
Adding the option `--if-exists` as well because otherwise we'd get error `role does not exist`

closes: #482


pkg/snapshot/generator/postgres/schema/pgdumprestore/snapshot_pg_dump_restore_generator.go
```diff
@@ -635,6 +644,40 @@ func (s *SnapshotGenerator) parseDump(d []byte) *dump {
...
+			 * ALTER TABLE .. DROP ..
+			 * DROP SCHEMA IF EXISTS pgstream;
+			 * DROP ...
+			 *
+			 * -- some comments
+			 * CREATE ...
+			 * -- the rest of the dump..
+			 *
+			 * Here we need to catch the part before CREATE statements begin.
+			 * First we look for "DROP SCHEMA IF EXISTS pgstream;" because it will always be there.
```
eminano: It won't always be there, only if all schemas are included. But usually, it will likely not be part of the snapshot (we don't actually want to have the pgstream state in the target db).

This way of splitting the cleanup part from the restore part seems a bit weak, prone to errors if something is out of place. One sturdier solution could be to do the dump with and without the clean, and using the diff as the clean part. That way we're not guessing which statements are part of the clean and which aren't.

These part needs tests 🙏

pkg/snapshot/generator/postgres/schema/pgdumprestore/snapshot_pg_dump_restore_generator.go
```diff
@@ -644,6 +687,7 @@ func (s *SnapshotGenerator) parseDump(d []byte) *dump {
 		indicesAndConstraints: []byte(indicesAndConstraints.String()),
 		sequences:             sequenceNames,
 		roles:                 roleNames,
+		cleanupPart:           []byte(cleanupPart.String()),
```
eminano: Right now the cleanup part is duplicated (once in the cleanup part, and another as part of the full dump).

pkg/snapshot/generator/postgres/schema/pgdumprestore/snapshot_pg_dump_restore_generator.go
```diff
@@ -719,3 +747,22 @@ func hasWildcardTable(tables []string) bool {
 func hasWildcardSchema(schemaTables map[string][]string) bool {
 	return schemaTables[wildcard] != nil
 }
+
+func getDumpsDiff(d1, d2 []byte) []byte {
```
eminano, line 752: A comment here to mention in which order the diff is performed would be helpful (returns all elements of d1 that are not on d2 for example).

pkg/snapshot/generator/postgres/schema/pgdumprestore/snapshot_pg_dump_restore_generator.go
eminano: As discussed, this will not cause errors since the clean is done with the `--if-exists` flag, so we'll leave it like this for now.

PR #497, opened 2025-09-03
Fix config data crash

eminano:
This PR fixes the issue where having the snapshot mode set to `full`/`data` would expect the `data` section of the yaml configuration to be present, or would crash otherwise. 

A check is added to make sure that if the data section is not provided, defaults will apply instead. 
Fuzzing has also been introduced, for local use for now, to identify crash paths for the yaml configuration parsing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,11 +15,11 @@ integration-test:
 .PHONY: fuzz
 fuzz:
 	@echo "Fuzzing YAML to Stream Config"
-	@go test -fuzz=FuzzYAMLToStreamConfig --fuzztime 5s github.com/xataio/pgstream/cmd/config
+	@go test -fuzz=FuzzYAMLToStreamConfig --fuzztime 30s github.com/xataio/pgstream/cmd/config
 	@echo "Fuzzing YAML Config Structure"
-	@go test -fuzz=FuzzYAMLConfigStructure --fuzztime 5s github.com/xataio/pgstream/cmd/config
+	@go test -fuzz=FuzzYAMLConfigStructure --fuzztime 30s github.com/xataio/pgstream/cmd/config
 	@echo "Fuzzing YAML Config Properties"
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -12,6 +12,15 @@ test:
 integration-test:
 	@PGSTREAM_INTEGRATION_TESTS=true go test -timeout 180s github.com/xataio/pgstream/pkg/stream/integration
 
+.PHONY: fuzz
+fuzz:
+	@echo "Fuzzing YAML to Stream Config"
+	@go test -fuzz=FuzzYAMLToStreamConfig --fuzztime 5s github.com/xataio/pgstream/cmd/config
```
eminano:
```suggestion
	@go test -fuzz=FuzzYAMLToStreamConfig --fuzztime 30s github.com/xataio/pgstream/cmd/config
```

Makefile
```diff
@@ -12,6 +12,15 @@ test:
...
+	@echo "Fuzzing YAML Config Structure"
+	@go test -fuzz=FuzzYAMLConfigStructure --fuzztime 5s github.com/xataio/pgstream/cmd/config
```
eminano:
```suggestion
	@go test -fuzz=FuzzYAMLConfigStructure --fuzztime 30s github.com/xataio/pgstream/cmd/config
```

Makefile
```diff
@@ -12,6 +12,15 @@ test:
...
+	@echo "Fuzzing YAML Config Properties"
+	@go test -fuzz=FuzzYAMLConfigProperties --fuzztime 5s github.com/xataio/pgstream/cmd/config
```
eminano:
```suggestion
	@go test -fuzz=FuzzYAMLConfigProperties --fuzztime 30s github.com/xataio/pgstream/cmd/config
```

PR #503, opened 2025-09-04
Add hstore to conn type map

agedemenli:
`COPY` with hstore fails with error `number of pairs (576791101) exceeds the maximum allowed (26843545)`. This PR fixes it by registering the type `hstore` to postgres connections and pools.

I wonder why we didn't face the same thing with `citext` though.

Types that are added by default:
https://github.com/jackc/pgx/blob/master/pgtype/pgtype_default.go

### DIFF
--- a/internal/postgres/pg_utils.go
+++ b/internal/postgres/pg_utils.go
@@ -108,14 +108,15 @@ func extractDatabase(url string) (string, error) {
 }
 
 func registerTypesToConnMap(ctx context.Context, conn *pgx.Conn) error {
-	var hstoreExists bool
-	err := conn.QueryRow(ctx, "SELECT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'hstore')").Scan(&hstoreExists)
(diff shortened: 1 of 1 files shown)


internal/postgres/pg_utils.go
```diff
@@ -105,3 +106,16 @@ func extractDatabase(url string) (string, error) {
 	}
 	return pgCfg.Database, nil
 }
+
+func registerTypesToConnMap(ctx context.Context, conn *pgx.Conn) error {
+	var hstoreExists bool
+	err := conn.QueryRow(ctx, "SELECT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'hstore')").Scan(&hstoreExists)
```
eminano: Would it not be simpler to just have one query to get the oid of the hstore type, and register the type if rows are found rather than performing two queries? If this is going to be done on every conn, we should try to make it as fast as possible 🏎️ .

eminano (approved):

PR #515, opened 2025-09-09
Add hstore transformer

agedemenli:
Adding hstore transformer, similar to json transformer. Hstore transformer executes a given list of operations on a hstore object. Each operation can either be `set` or `delete`. 
`delete` simply deletes the pair with the given key. If the key doesn't exist, it does nothing. If the parameter `error_not_exist` is set to `true`, errors out.
`set` operations can either have a value, or a template. In case it is a template, updating the value for the given key with the result of the template. It can be used with dynamic parameters and it supports all the functions supported by template and json transformers. In case `set` operation has a static value, it just assigns that value to the value for the specified key. In case the key does not exist, `set` operation adds it as a new value to the hstore object. However, again, if the parameter `error_not_exist` is set to true, it errors out.
The static value for a `set` operation can be nil. Since hstore supports null, we will be directly assigning the value for the given key to null in that case. However in template case, since template results are strings, it will not be nil. In case the result of a template is nil, it will be the string "<no value>". We can catch this result and assign the value to nil in that case. But I wonder what should we do in case the result of a template is actually the string "<no value>" 🤔


pkg/transformers/hstore_transformer.go
```diff
@@ -0,0 +1,220 @@
...
+		Parameters:     hstoreParams,
+	}
+}
+
+func getHstoreOperations(params ParameterValues) ([]*hstoreOperation, error) {
+	arrayAny, found, err := FindParameter[[]any](params, "operations")
+	if err != nil {
+		return nil, fmt.Errorf("operations must be an array: %w", err)
+	}
+	if !found {
```
eminano: You can probably collapse these two into one ( `if !found || len(arrayAny) == 0`), since they both should return `errOperationsMustBeProvided`.

eminano (approved):
Really nice! 🚀

PR #517, opened 2025-09-09
at_email transformer

spongenee:
We thought to contribute our custom email transformer that has 2 distinct features from `neosync_email` transformer:

- We can define domains to not anonymize with `exclude_domain`. In our case it's useful to not anonymize emails of our own company.
-  Similar to the `seed` parameter of `neosync_email`, it is deterministic with a random `salt` value, but for any input, including invalid ones. For the same input value, it will always be encrypted to the same output. This is useful because we might want to query our database with known customer emails.


docs/README.md
```diff
@@ -1306,6 +1306,48 @@ transformations:
               column: country_code
 ```
 
+</details>
+
+ <details>
+  <summary>at_email</summary>
```
eminano: We think this is a good opportunity to start a pgstream email custom transformer, since there isn't currently one (only `neosync` one available). So instead of `at_email` this transformer can just be called `email`. 
It will be the initial implementation, allowing us to extend/implement custom functionality whenever we need it, rather than relying on a third party. 

Thanks for helping us kick start it 🚀

pkg/transformers/at_email_transformer.go
```diff
@@ -0,0 +1,180 @@
...
+	"crypto/sha256"
+	"fmt"
+)
+
+var bytesOutputAlphabetLength = byte(len(bytesOutputAlphabet))
+var bytesKeep = []byte("',\\{}")
+var bytesOutputAlphabet = []byte("0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
+
+// Modifies `s` in-place.
+func ScrambleBytes(s []byte, salt string) []byte {
```
eminano: Since all these functions are used only by the transformers package, they don't need to be exported (keeping internals inside 😅 ).

pkg/transformers/at_email_transformer.go
```diff
@@ -0,0 +1,180 @@
...
+
+	}
+	atEmailCompatibleTypes = []SupportedDataType{
+		StringDataType,
+		CitextDataType,
+	}
+)
+
+func NewAtEmailTransformer(params ParameterValues) (*AtEmailTransformer, error) {
+	replacementDomain, err := FindParameterWithDefault(params, "replacement_domain", "@crypt.com")
```
eminano: I'd suggest a more obvious default for the `replacement_domain`, maybe `@scrambled.com`? 🍳

costi: The default should probably be example.com
sometimes the dev boxes start sending actual email and you don't want people at crypt.com or scramble.com to get those emails

pkg/transformers/at_email_transformer.go
spongenee: this is addressed in https://github.com/xataio/pgstream/pull/517/commits/baa413a57750f030f2d59add86a8fdbdb5b7c026

docs/README.md
spongenee: renamed in commit https://github.com/xataio/pgstream/pull/517/commits/b8eba923182d44063fa193355194ab2dfbb28b7e

spongenee:
@eminano thank you! I addressed the comments. Tests are good and pre-commit is no longer complaining. Feel free to make any further changes as you see fit. Thank you!

Issue #519, opened 2025-09-10
[ISSUE] Hstore will fail in an ugly way if key is missing.


seigel:
If you apply this patch to the hstore test you get an ugly error.  It can probably be corrected in a similar way to a patch that we will be submitting for the json store columns as well.  I am just keeping this hear to remind me.

```
diff --git a/pkg/transformers/hstore_transformer_test.go b/pkg/transformers/hstore_transformer_test.go
index a73efb5..c034ae2 100644
--- a/pkg/transformers/hstore_transformer_test.go
+++ b/pkg/transformers/hstore_transformer_test.go
@@ -377,6 +377,11 @@ func TestHstoreTransformer_Transform(t *testing.T) {
                                                "key":            "email",
                                                "value_template": "{{ masking \"email\" .GetValue }}",
                                        },
+                                       map[string]any{
+                                               "operation":      hstoreSetOpName,
+                                               "key":            "email_missing",
+                                               "value_template": "{{ masking \"email\" .GetValue }}",
+                                       },
                                        map[string]any{
                                                "operation":      hstoreSetOpName,
                                                "key":            "key3",
```

seigel:
Error:

```
                Error:          Received unexpected error:
                                hstore_transformer: cannot apply "set" operation[3] with key email_missing: error executing template: template: op[3] set email_missing:1:19: executing "op[3] set email_missing" at <.GetValue>: wrong type for value; expected string; got interface {}
```

Issue #518, opened 2025-09-10
[JSON Transformer] masking function breaks when source json does not contain path


spongenee:
We have been testing the json transformer and we have observed that custom functions breaks when source json does not contain path, resulting in data loss.  The target db will have no data for that column, other columns will be replicated as expected. 

The ideal behaviour is to skip if the path doesn't exist and move on to evaluate the next operation, so there will not be data loss. We can confirm `masking` function breaks but we believe it's the same for any of the template functions.

The following details what we observe:
With the following example config
```
        column_transformers:
          example_column:
            name: json
            parameters:
              operations:
                - operation: set
                  path: "name"
                  value_template: '"{{ masking "default" .GetValue }}"'
                  error_not_exist: true
                - operation: set
                  path: "email"
                  value_template: '"{{ masking "default" .GetValue }}"'
                  error_not_exist: true
                - operation: set
                  path: "passportId"
                  value_template: '"REDACTED"'
```

If source JSON is
```
{
  "name": "John Doe",
  "email": "john@doe.com"
}
```
Note `"passportId"` path doesn't exist, the target JSON will be as expected - both fields get redacted

```
{
  "name": "Jo***e",
  "email": "joh**com"
}
```

But if source JSON is 
```
{
  "name": "John Doe",
  "passportId": "A123456"
}
```
The target JSON will be null and the logs will show a dataloss ERR message, snapshot shows it's successful and replication continues.
```
transforming column error.message="cannot apply \"set\" operation[5] with path email: value by path \"email\" does not exist"
```
If I set `error_not_exist: false`, the only difference is the error message
```
"cannot apply \"set\" operation[5] with path settled_by_email: error executing template: template: op[5] set email:1:22: executing \"op[5] set email\" at <.GetValue>: wrong type for value; expected string; got interface {}"
```
But I expect the target JSON to be 
```
{
  "name": "Jo***e",
  "passportId": "REDACTED"
}
```

We are already working on a fix. But wanted to raise this issue and understand if this is known/expected behaviour.

PR #522, opened 2025-09-11
Add param skip_not_exist to hstore and json transformers

agedemenli:
Adding boolean operation parameter `skip_not_exist` to both hstore and json transformers. The default is true. When it's true, the operation will be skipped in case the key/path does not exist. The operation will go through if it's false.
Updating the docs and adding tests as well.

closes: #518 
closes: #519


pkg/transformers/hstore_transformer.go
```diff
@@ -93,6 +93,10 @@ func (t *HstoreTransformer) Transform(_ context.Context, value Value) (any, erro
 
 	transformed := toTransform
 	for idx, op := range t.operations {
+		t.hstoreVal.setValue(transformed, op.key)
+		if op.skipNotExist && !t.hstoreVal.exists {
```
eminano: We might want to deal with the `error_not_exists` case here, keeping similar logic together (what's done for the json transformer as well).

pkg/transformers/json_transformer_operation.go
```diff
@@ -33,31 +34,22 @@ func (o *jsonOperation) apply(inp []byte, jsonVal *jsonValue, buf *bytes.Buffer)
...
 			buf.Reset()
-			jsonVal.setValue(inp, o.path)
-			if o.errorNotExist && !jsonVal.exists {
-				return nil, fmt.Errorf("value by path \"%s\" does not exist", o.path)
-			}
 			if err = o.tmpl.Execute(buf, jsonVal); err != nil {
 				return nil, fmt.Errorf("error executing template: %w", err)
 			}
-			newValue := buf.Bytes()
-			res, err = sjson.SetRawBytesOptions(inp, o.path, newValue, jsonSetOpt)
```
eminano, line 46: You're replacing this `SetRawBytesOptions` by `SetBytesOptions`, is that intentional?

agedemenli: Yes it is. For some reason `SetRawBytesOptions` requires redundant quotation marks at the beginning and the end of the template. Please see the changes in the test file as an example. I've removed them with this PR, as they are not longer needed when using `SetBytesOptions`

Looking into it and saw that for []byte scenario the only difference between the functions is that it calls the internal `set` function with `stringify = true` 🤔

PR #530, opened 2025-09-17
[Snapshots] Expose no privileges option

eminano:
This PR updates the pgdump/restore snapshot generator to expose the `--no-privileges` flag in the pgstream configuration. This will allow to avoid privilege related commands in the dump.


agedemenli (approved):
thank you <3

PR #543, opened 2025-09-24
Support special characters in Postgres password

eminano:
#### Description

This PR updates the connection string parsing so that special characters are supported in Postgres URLs.

#### Type of Change

#### Changes Made

- Added parsing of URL when connection fails to make sure the password is encoded
- Added a `ParseConfig` util in the pglib to make sure all calls to parse config apply the encoding
- Fixed search path appending in init so it works both with empty/existing parameters

#### Testing

- [X] All existing tests pass

### DIFF
--- a/internal/postgres/pg_utils.go
+++ b/internal/postgres/pg_utils.go
@@ -209,15 +209,16 @@ func escapeConnectionURL(rawURL string) (string, error) {
 	userInfo := matches[2]    // "username:password"
 	hostAndPath := matches[3] // "host:port/database?params"
 
-	// Find the last colon in userInfo to split username and password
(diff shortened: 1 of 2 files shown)


agedemenli:
What if the password has a colon in it? 
I'd expect `postgres://user:p:ss@localhost:5432/mydb` to become `postgres://user:p%3Ass@localhost:5432/mydb`
But this implementation assumes that the last colon in the url is the separator between username and password, as if a password cannot have a colon in it.
I know it is ambiguous, we cannot know which one is the separator. Am I missing something?

agedemenli:
Similarly, what should be do if the password has an `@`?

PR #559, opened 2025-10-11
Add option for ignoring processing errors

tsg:
#### Description

This adds two config options for the snapshot configuration:

* `ignore_row_processing_errors` (`PGSTREAM_POSTGRES_SNAPSHOT_IGNORE_ROW_PROCESSING_ERRORS`) - if set to true, it will continue the snapshot in case there are errors on a particular row / batch.
* `log_row_on_error" (`PGSTREAM_POSTGRES_SNAPSHOT_LOG_ROW_ON_ERROR`) - if set to true, it will log the row (including user data) before continuing / exiting.

Both of these options are meant for troubleshooting and should be used with care in production environments:
*  Ignoring errors can result in consistency issues, so while the immediate error is ignored, foreign-key or other constraints might fail later in the process.
* Logging the user data might leak PII into the logs.

#### Type of Change

- [x] 📚 Documentation update
- [ ] 🔨 Build/CI changes
- [ ] 🧹 Code cleanup

#### Testing

- [ ] All existing tests pass

#### Checklist

- [ ] Self-review completed
- [ ] Code is well-commented


pkg/snapshot/generator/postgres/data/pg_snapshot_generator.go
```diff
@@ -357,11 +363,32 @@ func (sg *SnapshotGenerator) snapshotTableRange(ctx context.Context, snapshotID
...
+						"schema": table.schema,
+						"table":  table.name,
+						"error":  err.Error(),
+					}
+					if sg.logRowOnError {
+						if rowJSON, jsonErr := jsonlib.Marshal(row); jsonErr == nil {
+							logFields["row"] = string(rowJSON)
+						}
+					}
+
```
eminano, line 383: Maybe we could simplify this by always logging the error.
```
if sg.logRowOnError {
	if rowJSON, jsonErr := jsonlib.Marshal(row); jsonErr == nil {
			logFields["row"] = string(rowJSON)
	}
}
  
sg.logger.Error(err, "failed to process snapshot row", logFields)

if sg.ignoreRowProcessingErrors {
	continue
}

return fmt.Errorf("processing snapshot row: %w", err)
```

config_template.yaml
```diff
@@ -22,6 +22,8 @@ source:
         schema_workers: 4 # number of schema tables to be snapshotted in parallel. Defaults to 4
         table_workers: 4 # number of workers to snapshot a table in parallel. Defaults to 4
         batch_bytes: 83886080 # bytes to read per batch (defaults to 80MiB)
+        ignore_row_processing_errors: false # if true, log errors during row processing and continue. Defaults to false. Warning: can result in consistency errors.
```
eminano, line 25: We need to update the configuration docs to include the new options as well.
https://github.com/xataio/pgstream/blob/main/docs/configuration.md#%EF%B8%8F-configuration

PR #562, opened 2025-10-14
[Snapshot] Change role snapshot default mode

eminano:
#### Description

This PR updates the role snapshot mode to default to `no_passwords` instead of `enabled`. Password snapshotting requires an extra level of permissions (to be able to access `pg_authid` table), and could present a leak. The default should be for roles to have different passwords in different databases for security purposes.

#### Type of Change

- [X] 📚 Documentation update


docs/configuration.md
```diff
@@ -178,7 +178,7 @@ Here's a list of all the environment variables that can be used to configure the
...
-| PGSTREAM_POSTGRES_SNAPSHOT_ROLES_SNAPSHOT_MODE       | "enabled"                    | No       | When using `pg_dump`/`pg_restore` to snapshot schema for Postgres targets, option to enable/disable snapshotting roles. Could be set to "no_passwords" to snapshot roles, without passwords.                                                                                                                 |
+| PGSTREAM_POSTGRES_SNAPSHOT_ROLES_SNAPSHOT_MODE       | "no_passwords"               | No       | When using `pg_dump`/`pg_restore` to snapshot schema for Postgres targets, option to enable/disable snapshotting roles. Can be set to "no_passwords" to snapshot roles, without passwords.                                                                                                                   |
```
exekias: having that this is the default now, maybe it's not entirely clear how to change the behavior away from `no_passwords`?

eminano:
```suggestion
| PGSTREAM_POSTGRES_SNAPSHOT_ROLES_SNAPSHOT_MODE       | "no_passwords"               | No       | When using `pg_dump`/`pg_restore` to snapshot schema for Postgres targets, controls how roles are snapshotted. Possible values: "enabled" (snapshot all roles including passwords), "disabled" (do not snapshot roles), "no_passwords" (snapshot roles but exclude passwords).                                                                                                                |
```

Issue #506, opened 2025-09-04
Apply reconnection backoff to postgres connections


eminano:
If a connection is dropped, attempt to reconnect for a period of time before giving up.

PR #583, opened 2025-11-04
Replication reconnection support

eminano:
#### Description

This PR updates the replication handler to support retries when receiving replication messages or when syncing the LSN. This means pgstream will apply a configurable retry policy (either exponential or constant backoff) to the replication connection, allowing for temporary connection issues without disrupting the pgstream service, which will be more sturdy against connection errors.

##### Related Issue(s)

- Related to #506 

#### Type of Change

- [X] 📚 Documentation update

#### Changes Made

- Added a replication handler retrier layer to apply backoff
- Add retry policy configuration to the source postgres
- Add ability to reset connection to the replication handler

#### Testing

- [X] All existing tests pass


exekias (approved):
Really good improvement! I'm now wondering, what happens in the case of a disconnect to the postgres we are writing to? do we need a retry mechanism there?

Issue #584, opened 2025-11-06
[Snapshot] Process gets stuck


eminano:
**Describe the bug**
pgstream snapshot hangs after a period of time, making no apparent progress. No logs, or metrics are reported, but the program is still runinng, pointing to a deadlock somewhere in the process (either pgstream or source/target databases).

**Expected behavior**

**Screenshots**

**Setup (please complete the following information):**
 - pgstream version: v0.8.7
 - Postgres version: ?
 - Postgres environment: GCP CloudSQL source, Xata database target

PR #585, opened 2025-11-06
[Snapshot] Cache custom type oid check

eminano:
#### Description

This PR updates the postgres type mapper to add a caching layer for oids not present in the pgtype.Map which contains PostgreSQL oids for common types. This means that rather than having to perform a query each time we receive an unknown oid, we save the type in the cache so that it's available for the next checks. This is particularly helpful during snapshots, which perform a type check for each row they read, reducing the number of connections required when performing highly concurrent snapshots.

This could lead to connection pool saturation.

##### Related Issue(s)

- Related to #584 

#### Type of Change

#### Changes Made

- Updated the existing `StringMap` in the sync library to work with any comparable types, not only strings as keys.
- Add a cache to the postgres oid type mapping flow

#### Testing

- [X] All existing tests pass

### DIFF
--- a/internal/postgres/pg_mapper.go
+++ b/internal/postgres/pg_mapper.go
@@ -11,12 +11,24 @@ import (
 	"github.com/jackc/pgx/v5/pgtype"
 )
 
+// Mapper provides PostgreSQL type information mapping from OIDs to type names.
+// It uses a combination of the pgx type map for standard types and a custom
+// cache for user-defined types, querying the database when necessary.
 type Mapper struct {
-	querier      Querier
-	pgMap        *pgtype.Map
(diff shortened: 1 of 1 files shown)


internal/postgres/pg_mapper.go
```diff
@@ -6,18 +6,22 @@ import (
...
+
 	"github.com/jackc/pgx/v5/pgtype"
 )
 
 type Mapper struct {
-	querier Querier
-	pgMap   *pgtype.Map
+	querier      Querier
+	pgMap        *pgtype.Map
+	customOIDMap *synclib.Map[uint32, string]
```
exekias, line 26: should we make it clear through the name or a comment that this is a cache?

exekias (approved):

PR #588, opened 2025-11-08
Fix ordering of DROP TRIGGER statements in the filtered schema

tsg:
#### Description

In a similar issue with https://github.com/xataio/pgstream/pull/573 we're seeing error of this form:

```
ERROR: relation "public.table" does not exist
```

Which seem to come from DROP TRIGGER statmenets in the filtered schema like:

```
DROP TRIGGER IF EXISTS trigger_name ON public.table;
```

because it's called before `public.table` is created.

##### Related Issue(s)

- Fixes #(issue number)
- Closes #(issue number)
- Related to #(issue number)

#### Type of Change

- [ ] 📚 Documentation update
- [ ] 🔨 Build/CI changes
- [ ] 🧹 Code cleanup

#### Changes Made

-
-
-

#### Testing

- [ ] All existing tests pass

#### Checklist

- [ ] Self-review completed
- [ ] Code is well-commented


eminano:
This shouldn't happen. The drop statements are coming from the clean stage of the dump. The order should be valid unless the clean is enabled and the table doesn't exist, but even in that case, it shouldn't be preventing the dump, those errors should be ignored. 

Can you share more details about what triggers this?

Issue #123, opened 2025-01-24
Update trigger function to detect new DDL events


eminano:
Currently we only update the schema log view when receiving create/alter table and drop table/schema DDL commands. We should update it to include the newly added constraints/indices commands.

https://github.com/xataio/pgstream/blob/main/migrations/postgres/5_create_pgstream_log_schema_function.up.sql

Issue #116, opened 2025-01-21
Add unique constraints to schema log


eminano:
Keep track of unique constraints to the schema log view.

Example from pgroll schema view: https://github.com/xataio/pgroll/blob/main/pkg/state/init.sql#L220

Issue #115, opened 2025-01-21
Add foreign keys to schema log


eminano:
Keep track of foreign keys in the schema log view.

Example from pgroll schema view: https://github.com/xataio/pgroll/blob/main/pkg/state/init.sql#L230

Issue #114, opened 2025-01-21
Add check constraints to schema log


eminano:
Add check constraints to the schema log view.

Example can be found in pgroll schema view: https://github.com/xataio/pgroll/blob/main/pkg/state/init.sql#L210

PR #609, opened 2025-11-13
Replicate indexes, constraints, and FKs from schema logs

Arochka:
#### Description

  Makes schema replication fully index/constraint/foreign-key aware so downstream targets stay aligned without manual fixes. Schema diffing now canonicalizes CREATE INDEX definitions, tracks add/remove/rename events keyed by name, filters constraint-backed indexes, and keeps constraint/FK definitions in sync. The Postgres WAL DDL adapter consumes that richer diff to emit
  CREATE/DROP/RENAME INDEX plus ADD/DROP CONSTRAINT and ADD/DROP FK statements—with IF NOT EXISTS guards—keeping replayed tables, indexes, constraints, and foreign keys consistent.

  ##### Related Issue(s)
  - Related to #113, #114, #115, #116, #123

  #### Type of Change

  - [ ] 📚 Documentation update
  - [ ] 🔨 Build/CI changes
  - [ ] 🧹 Code cleanup

  #### Changes Made

- Extend schemalog models/diffing to normalize index definitions, track add/remove/rename events, and include constraint/FK metadata so replicas know exactly what to replay while skipping indexes created implicitly by constraints.
  - Update the Postgres WAL DDL adapter to emit DROP/CREATE/RENAME INDEX, ADD/DROP CONSTRAINT, and ADD/DROP FOREIGN KEY statements, all derived from the schema diff and guarded with IF NOT EXISTS where appropriate.
  - Expand unit tests in `pkg/schemalog` and `pkg/wal/processor/postgres` to cover index rename detection, constraint/FK propagation, constraint-backed index handling, and the generated DDL.

  #### Testing

  - [x] Unit tests added/updated (go test ./pkg/schemalog ./pkg/wal/processor/postgres)
  - [x ] Integration tests added/updated
  - [x] All existing tests pass

  #### Checklist

  - [x] Code follows project style guidelines (pre-commit run --all-files)
  - [x] Self-review completed
  - [ ] Code is well-commented


migrations/postgres/9_add_index_replication_support.up.sql
```diff
@@ -0,0 +1,199 @@
+-- Migration: Add full index replication support (triggers, log_schema, get_schema)
```
eminano, line 1: You need to run `make gen-migrations` so that your migrations are added to the bindata that is later applied by the init/destroy migrator. 
https://github.com/xataio/pgstream/blob/main/migrations/postgres/migrations.go

pkg/schemalog/schema_diff.go
```diff
@@ -20,6 +18,8 @@ type TableDiff struct {
 	ColumnsAdded          []Column
 	ColumnsRemoved        []Column
 	ColumnsChanged        []ColumnDiff
+	IndexesAdded          []Index
+	IndexesRemoved        []Index
```
eminano, line 25: Right now you're not tracking indexes changes (`ALTER INDEX`). You have handling for name changes by dropping/creating a new index with the new name, but this might not be the most performant approach with very large indexes, where applying the `ALTER` statement might be a better solution. 

It can stay like this for now, but it's worth noting it as a limitation of the implemented DDL index support. We can open a separate issue to handle `ALTER INDEX` if necessary (you're already tracking it on the schema log with this PR).

pkg/wal/processor/postgres/postgres_wal_ddl_adapter.go
```diff
@@ -252,3 +292,11 @@ func (a *ddlAdapter) newDDLQuery(schema, table, sql string) *query {
 		isDDL:  true,
 	}
 }
+
+func buildDropIndexQuery(schemaName, indexName string) string {
+	if indexName == "" {
+		return ""
+	}
+	qualified := fmt.Sprintf("%s.%s", pglib.QuoteIdentifier(schemaName), pglib.QuoteIdentifier(indexName))
```
eminano: There's a util in the pglib you can use. It's `QuoteQualifiedIdentifier`.

Arochka:
Thanks a lot for the feedback!

I’ve just added support for constraints and foreign keys.
I also ran the gen-migrations step, which I had forgotten to commit earlier.

Thanks as well for the ALTER INDEX suggestion — I’ve made some updates, and it should indeed help improve performance.

pkg/schemalog/schema_diff.go
```diff
@@ -150,6 +163,50 @@ func computeTableDiff(old, new *Table) *TableDiff {
 		}
 	}
 
+	matchRenamedIndexes(diff)
```
eminano: Would it make sense to deal with renames (and other `ALTER INDEX` operations) by checking during the equality comparison on line 154? We could just check if they're not equal and if the description starts with `ALTER INDEX` then we can add it as a `IndexesChanged` with the description sql to be applied directly. 

I think it would remove all the complexity from the `matchRenamedIndexes` and also remove the need for crafting the sql in the [ddl adapter](https://github.com/xataio/pgstream/pull/609/commits/685a2221468d4d6474a0dec2c2df77280277fd67#diff-f44975e58ccf9594a166631b719cfdc93917eef008eef55f0133f386d7bda887R443). It would also mean we don't only cover renames, but other alter index operations as well.

pkg/stream/integration/pg_pg_integration_test.go
```diff
@@ -303,6 +303,114 @@ func Test_PostgresToPostgres_Snapshot(t *testing.T) {
 	}
 }
 
+func Test_PostgresToPostgres_SchemaObjects(t *testing.T) {
```
eminano, line 306: Nice! 🙌

pkg/wal/processor/postgres/postgres_wal_ddl_adapter.go
```diff
@@ -111,6 +120,29 @@ func ensureIndexHasIfNotExists(definition string) string {
...
+		if constraintCreatesIndex(constraint.Type) && constraint.Name != "" {
+			indexes[constraint.Name] = struct{}{}
+		}
+	}
+	return indexes
+}
+
+func constraintCreatesIndex(constraintType string) bool {
+	switch {
+	case strings.EqualFold(constraintType, "UNIQUE"):
```
eminano: Just as a note, you can collapse this into a single case in go:
```
switch {
	case strings.EqualFold(constraintType, "UNIQUE"),
	strings.EqualFold(constraintType, "PRIMARY KEY"),
	strings.EqualFold(constraintType, "EXCLUDE"):
	   return true
    default:
       return false
  }

pkg/schemalog/schema_diff.go
Arochka: Great call, thanks for the suggestion!

PR #623, opened 2025-11-20
Fix identity column replication

### LINKED ISSUE #430: [Postgres Processor] Error updating `GENERATED ALWAYS AS IDENTITY` columns
Updating a column of type `GENERATED ALWAYS AS IDENTITY` throws the error: `column can only be updated to DEFAULT`.

This is because we're trying to set that column, but due to its type it can only be updated with `DEFAULT`. 

When we insert data we use the `OVERRIDING SYSTEM VALUE ` which resolves this problem, but for UPDATE there's no such thing.

We can fix it by flagging `GENERATED ALWAYS AS IDENTITY` columns, and using `DEFAULT` for those instead of inserting the value from the source, but that would mean the source and the target would potentially fall out of sync due to the different sequences (ie., the target has inserts that the source doesn't).

If we want to keep it consistent with the source, then maybe the `GENERATED ALWAYS AS IDENTITY` should be `GENERATED BY DEFAULT AS IDENTITY` so that it can be set from the source, but it makes the schema out of sync with the source.

We'll need to decide on which consistency we want to prioritise during replication, data or schema.

A workaround until this is fixed is to use `GENERATED BY DEFAULT AS IDENTITY` instead of `GENERATED ALWAYS AS IDENTITY` in the target database.

eminano:
#### Description

This PR updates the schemalog tracking to be able to identify identity columns, and also the generated column kind. In order to avoid breaking changes, the generated kind flag is added to the existing generated boolean flag. 

The identity information is then used as part of the generated column handling, which was getting skipped before, since identity columns are not flagged as generated. 

##### Related Issue(s)

- Fixes #430 

#### Type of Change

#### Changes Made

- Add migration to track identity columns as well as the kind for generated columns
- Update the schemalog type
- Include identity columns as part of the generated checks and special handling
- Use the new column information to build accurate column definitions in the ddl adapter
- Improve postgres error mapping

#### Testing

- [X] All existing tests pass

### DIFF
--- a/internal/postgres/errors.go
+++ b/internal/postgres/errors.go
@@ -64,6 +64,14 @@ func (e *ErrRelationAlreadyExists) Error() string {
 	return fmt.Sprintf("relation already exists: %v", e.Details)
 }
 
+type ErrRuleViolation struct {
+	Details string
+}
+
+func (e *ErrRuleViolation) Error() string {
+	return fmt.Sprintf("rule violation: %s", e.Details)
+}
+
 func MapError(err error) error {
 	if pgconn.Timeout(err) {
 		return fmt.Errorf("%w: %w", ErrConnTimeout, err)
@@ -89,6 +97,11 @@ func MapError(err error) error {
 			// 42000 	syntax_error_or_access_rule_violation
 			// 42601 	syntax_error
 			return &ErrSyntaxError{
+				Details: pgErr.Message,
+			}
+		case "428C9":
+			// 428C9 	generated_always
+			return &ErrRuleViolation{
 				Details: pgErr.Message,
 			}
 		case "42501":
--- a/internal/postgres/retrier/pg_querier_retrier.go
+++ b/internal/postgres/retrier/pg_querier_retrier.go
@@ -157,10 +157,16 @@ func (q *Querier) isRetriableError(err error) bool {
 	permissionDenied := &postgres.ErrPermissionDenied{}
 	constraintViolation := &postgres.ErrConstraintViolation{}
 	alreadyExists := &postgres.ErrRelationAlreadyExists{}
+	ruleViolation := &postgres.ErrRuleViolation{}
+	syntaxError := &postgres.ErrSyntaxError{}
+	doesNotExist := &postgres.ErrRelationDoesNotExist{}
 	switch {
 	case errors.As(mappedErr, &permissionDenied),
 		errors.As(mappedErr, &constraintViolation),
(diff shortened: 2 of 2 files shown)


migrations/postgres/12_update_column_flags_in_get_schema_function.up.sql
```diff
@@ -0,0 +1,166 @@
...
+        table_oids.table_name AS table_name,
+        table_oids.table_oid AS table_oid,
+        table_oids.table_pgs_id AS table_pgs_id,
+        format('%s-%s', table_oids.table_pgs_id, pg_attribute.attnum) AS column_pgs_id,
+        pg_attribute.attname AS column_name,
+        format_type(pg_attribute.atttypid, pg_attribute.atttypmod) AS column_type,
+        pg_get_expr(pg_attrdef.adbin, pg_attrdef.adrelid) AS column_default,
+        NOT (pg_attribute.attnotnull OR pg_type.typtype = 'd' AND pg_type.typnotnull) AS column_nullable,
+        NOT ( pg_attribute.attgenerated = '') AS column_generated,
+        pg_attribute.attgenerated AS column_generated_kind,
```
eminano, line 37: Line added

migrations/postgres/12_update_column_flags_in_get_schema_function.up.sql
```diff
@@ -0,0 +1,166 @@
...
+        pg_attribute.attidentity AS column_identity,
```
eminano, line 38: Line added

migrations/postgres/12_update_column_flags_in_get_schema_function.up.sql
```diff
@@ -0,0 +1,166 @@
...
+        columns.table_oid,
+        columns.table_pgs_id AS table_pgs_id,
+        jsonb_agg(jsonb_build_object(
+            'pgstream_id', columns.column_pgs_id,
+            'name', columns.column_name,
+            'type', columns.column_type,
+            'default', columns.column_default,
+            'nullable', columns.column_nullable,
+            'generated', columns.column_generated,
+            'generated_kind', columns.column_generated_kind,
```
eminano, line 112: Line added

migrations/postgres/12_update_column_flags_in_get_schema_function.up.sql
```diff
@@ -0,0 +1,166 @@
...
+            'identity', columns.column_identity,
```
eminano, line 113: Line added

Issue #645, opened 2025-12-11
[Replication] Sequence support


eminano:
**Is your feature request related to a problem? Please describe.**
Sequences are not currently supported during replication. During the snapshot they are created correctly in the target (including the sequence data), but during replication:
- The sequence data is not incremented when data is inserted, since the sequence value in the row is manually set by the source 
- When a column that uses a sequence is created, the sequence backing it is not created in the target

This can lead to inconsistencies between the source and the target. At the moment, there's data fidelity but no schema fidelity for sequences.

**Describe the solution you'd like**
- For sequences to be incremented when an insert comes from the source during replication
- Sequences to be created when a new column that uses them is created during replication

**Describe alternatives you've considered**
- Running schema only snapshot for sequence sync
- Updating the sequence in the target manually

PR #648, opened 2025-12-12
[Replication] Sequence support (implicit)

eminano:
#### Description

This PR adds sequence support during replication, aiming to achieve both data and schema fidelity. Until now, only data fidelity was achieved. With the changes in this PR, the sequence value is updated after inserts, and columns with sequences (ie., `SERIAL`) are created properly instead of ignoring the sequence in the target.

##### Related Issue(s)

- Relates to #645 

#### Type of Change

#### Changes Made

- Add column sequences to the schema observer
- Use column sequences to set the sequence value with `setval` after inserts
- Support `SERIAL`, `BIGSERIAL` and `SMALLSERIAL` in column definition during DDL processing

#### Testing

- [X] All existing tests pass

#### Additional Notes

This PR does not add support for creating sequences explicitly (`CREATE SEQUENCE`). That will be handled in a separate PR.

### DIFF
--- a/pkg/wal/processor/postgres/postgres_wal_dml_adapter_test.go
+++ b/pkg/wal/processor/postgres/postgres_wal_dml_adapter_test.go
@@ -197,6 +197,35 @@ func TestDMLAdapter_walDataToQueries(t *testing.T) {
 			},
 		},
 		{
+			name: "insert with sequences - for copy enabled",
+			walData: &wal.Data{
+				Action: "I",
+				Schema: testSchema,
+				Table:  testTable,
+				Columns: []wal.Column{
+					{ID: columnID(1), Name: "id", Value: float64(1)},
+					{ID: columnID(2), Name: "name", Value: "alice"},
+				},
+				Metadata: wal.Metadata{
+					InternalColIDs: []string{columnID(1)},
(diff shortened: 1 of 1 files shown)


pkg/wal/processor/postgres/postgres_writer.go
```diff
@@ -56,7 +56,9 @@ func newWriter(ctx context.Context, config *Config, schemaLogStore schemalog.Sto
 		return nil, err
 	}
 
-	w.adapter, err = newAdapter(ctx, schemaLogStore, w.logger, config.URL, config.OnConflictAction, config.BulkIngestEnabled)
+	forCopy := writerType == bulkIngestWriter
```
exekias, line 59: is this change related?

eminano: This wasn't working as expected, it was using the BulkIngestEnabled flag, which was incorrect. I discovered it while testing the sequence, since it uses the `forCopy` in its logic, but otherwise it could be a separate PR.

exekias: that's fine, was just making sure it was not a leftover!

PR #681, opened 2026-01-19
Fix double encoding in postgres password

### LINKED ISSUE #680: Authentication error when Postgres password contains escaped characters
**Describe the bug**
When a Postgres password contains escaped characters, authentication fails due to double encoding.

**To Reproduce**

1. Use a postgres user with special characters in the password, and escape them manually when providing it in the configuration.
2. Run pgstream 
4. See authentication error

**Expected behavior**
Escaped characters to not be double encoded.

eminano:
#### Description

This PR updates the password escaping logic to handle passwords that are already escaped, and prevent double encoding.

##### Related Issue(s)

- Fixes #680 

#### Type of Change

#### Testing

- [X] All existing tests pass

### DIFF
--- a/internal/postgres/pg_utils.go
+++ b/internal/postgres/pg_utils.go
@@ -237,8 +237,17 @@ func escapeConnectionURL(rawURL string) (string, error) {
 	if username == "" {
 		return "", errInvalidURL
 	}
+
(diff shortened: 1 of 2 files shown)


exekias:
I would expect some library to exist for this job? still LGTM

PR #685, opened 2026-01-20
fix: pre-serialize JSONB columns with Sonic to prevent encoding misma…

VikramChennai:
When pgx receives a map[string]any for JSONB columns, it re-serializes using encoding/json, which can produce different output than Sonic (used to parse wal2json). This mismatch causes 'invalid input syntax for type json' errors on the target database for complex JSONB with Unicode, emojis, or special escapes.

Fix: Pre-serialize JSONB/JSON columns to []byte using Sonic in filterRowColumns() before passing to pgx. This ensures consistent encoding throughout the pipeline.

This also cascades a mess of errors for the entire row being rejected resulting in jsonB errors + others that are just false flags

### DIFF
--- a/pkg/wal/processor/postgres/jsonb_serialization_test.go
+++ b/pkg/wal/processor/postgres/jsonb_serialization_test.go
@@ -80,3 +80,30 @@ func TestFilterRowColumnsJSONBArrayHandling(t *testing.T) {
 	require.Len(t, parsed, 2)
 	require.Equal(t, "🎉", parsed[0].(map[string]any)["emoji"])
 }
+
+func TestBuildWhereQueryJSONBHandling(t *testing.T) {
+	t.Parallel()
+
+	// Simulates REPLICA IDENTITY FULL with JSONB column in identity
+	var jsonbValue map[string]any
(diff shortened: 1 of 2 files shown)


pkg/wal/processor/postgres/jsonb_serialization_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	_, values := (&dmlAdapter{}).filterRowColumns(cols, schemaInfo{})
+
+	jsonbResult, ok := values[1].([]byte)
+	require.True(t, ok, "JSONB array should be pre-serialized to []byte, got %T", values[1])
+
+	var parsed []any
+	require.NoError(t, json.Unmarshal(jsonbResult, &parsed))
+	require.Len(t, parsed, 2)
+	require.Equal(t, "🎉", parsed[0].(map[string]any)["emoji"])
```
eminano, line 81:
```suggestion
	emoji, ok := parsed[0].(map[string]any)["emoji"]
	require.True(t, ok, "first array item should have emoji field")
	require.Equal(t, "🎉", emoji)
```

PR #710, opened 2026-02-03
Update release notes

eminano:
Release notes preview:
https://github.com/xataio/pgstream/blob/update-release-notes/docs/releases/RELEASE_NOTES_v1.md

### DIFF
--- a/docs/releases/RELEASE_NOTES_v1.md
+++ b/docs/releases/RELEASE_NOTES_v1.md
@@ -1,9 +1,10 @@
 # pgstream v1.0.0 Release Notes
 
-This is a **major, breaking release**.
+**Release Date:** February 4th, 2026
 
-pgstream v1.0.0 introduces a new **stateless DDL replication architecture**.
-The legacy schema log mechanism has been removed, and schema changes are now replicated directly from PostgreSQL WAL.
+## 🎉 Major Release: Stateless DDL Replication
+
+This is a **major breaking release** that fundamentally transforms how pgstream handles DDL (Data Definition Language) replication. Version 1.0.0 introduces a stateless architecture that eliminates the need for schema log tables and provides a more robust, maintainable solution for tracking database schema changes.
 
 ## Table of Contents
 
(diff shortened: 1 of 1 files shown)


docs/img/pgstream_v1_schema_replication.png
gulcin, line 1: Love the addition of the image, it will be super useful!

docs/releases/RELEASE_NOTES_v1.md
```diff
@@ -1,342 +1,225 @@
 # pgstream v1.0.0 Release Notes
 
-**Release Date:** February 4th, 2026
+This is a **major, breaking release**.
```
gulcin: We could keep the release date maybe? Would not it be useful?

docs/releases/RELEASE_NOTES_v1.md
```diff
@@ -1,342 +1,225 @@
...
 
-## 🎉 Major Release: Stateless DDL Replication
+pgstream v1.0.0 introduces a new **stateless DDL replication architecture**.
```
gulcin: Also the title here can stay imo with the edited text below.

docs/releases/RELEASE_NOTES_v1.md
```diff
@@ -1,342 +1,225 @@
...
-# Option 3: Run only database migrations (without creating replication slot)
-pgstream init --config config.yaml --migrations-only
 ```
 
-The system will automatically install the appropriate migrations based on your configuration.
-
-**Note:** The `--migrations-only` flag runs only the database migrations (creating the pgstream schema, tables, functions, and triggers) without creating the replication slot. This is useful when you want to set up the schema separately or when using different database credentials for migrations versus replication.
-
-### For Existing Installations
+#### Install migrations only (no replication slot)
```
gulcin, line 96: This might need a bit more info to clarify what does `--migrations-only` flag do. I think the removed bit above can be added here.

docs/releases/RELEASE_NOTES_v1.md
```diff
@@ -1,342 +1,225 @@
...
-   # Stop running pgstream processes
-   ```
+Upgrading requires re-initialization.
 
-2. **Clean up old pgstream state** (recommended)
+Recommended steps:
 
-   v1.0.0 initialization will not automatically remove old v0.x migrations. To clean up the old state, use `pgstream destroy` before upgrading:
+1. Stop the running pgstream instance
+2. Remove existing state
```
gulcin, line 113: How can one remove the existing state? 

I think for each step giving the matching commands would be helpful.

gulcin (approved):

docs/releases/RELEASE_NOTES_v1.md
eminano: That's the command below. Do you think it needs more?

PR #724, opened 2026-02-13
[Replication] Fix duplicate alter drop events

eminano:
#### Description

This PR updates the `pgstream.emit_ddl` function to prevent sending duplicate events for `ALTER ... DROP` queries. Currently there are event triggers for both **ddl_command_end** and **sql_drop**.

The reason they both exist is because they provide different information:

- **sql_drop**: Needed for explicit DROP commands (DROP TABLE, DROP INDEX, etc.) because `pg_event_trigger_ddl_commands` returns nothing after the objects are deleted. Instead, `pg_event_trigger_dropped_objects` is used in this case.
- **ddl_command_end**: Handles CREATE/ALTER operations where objects still exist and metadata can be queried

For `ALTER ... DROP` statements, both emit the same DDL event but with different objects metadata. This creates duplication during replication.

This PR makes **sql_drop** only emit for true `DROP` commands, not `ALTER` commands, since for those the metadata is already available in the `ddl_command_end` trigger.

#### Type of Change

#### Changes Made

- Updated core and injector migrator files to skip non `DROP` statements for `sql_drop` trigger

#### Testing

- [X] All existing tests pass

### DIFF
--- a/pkg/stream/stream_status_checker.go
+++ b/pkg/stream/stream_status_checker.go
@@ -29,7 +29,7 @@ type StatusChecker struct {
 	ruleValidatorBuilder func(context.Context, string, []string) (ruleValidator, error)
 }
 
-type ruleValidator func(ctx context.Context, rules transformer.Rules) (map[string]transformer.ColumnTransformers, error)
+type ruleValidator func(ctx context.Context, rules transformer.Rules) (*transformer.TransformerMap, error)
 
 type migrator interface {
 	Status() ([]migratorlib.MigrationStatus, error)
--- a/pkg/stream/stream_status_checker_test.go
+++ b/pkg/stream/stream_status_checker_test.go
@@ -112,7 +112,7 @@ func TestStatusChecker_Status(t *testing.T) {
 	}
 
 	validRuleValidatorBuilder := func(context.Context, string, []string) (ruleValidator, error) {
-		return func(ctx context.Context, rules transformer.Rules) (map[string]transformer.ColumnTransformers, error) {
+		return func(ctx context.Context, rules transformer.Rules) (*transformer.TransformerMap, error) {
 			return nil, nil
 		}, nil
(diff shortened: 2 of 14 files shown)


migrations/postgres/core/3_fix_duplicate_ddl_emission.up.sql
```diff
@@ -0,0 +1,129 @@
...
+
+    -- Capture the actual DDL command being executed
+    ddl_command := current_query();
+
+    -- Determine which schema this affects and collect all affected objects
+    IF tg_event = 'sql_drop' THEN
+        -- For sql_drop, only emit for true DROP commands (DROP TABLE, DROP INDEX, etc.)
+        -- Skip ALTER commands that happen to drop something (like ALTER TABLE DROP COLUMN)
+        -- because ddl_command_end will handle those and provide better metadata
+        IF tg_tag NOT LIKE 'DROP %' THEN
```
eminano, line 33: This if check is the change introduced in this PR.

migrations/postgres/injector/4_fix_duplicate_ddl_emission.up.sql
```diff
@@ -0,0 +1,151 @@
...
+
+    -- Capture the actual DDL command being executed
+    ddl_command := current_query();
+
+    -- Determine which schema this affects and collect all affected objects
+    IF tg_event = 'sql_drop' THEN
+        -- For sql_drop, only emit for true DROP commands (DROP TABLE, DROP INDEX, etc.)
+        -- Skip ALTER commands that happen to drop something (like ALTER TABLE DROP COLUMN)
+        -- because ddl_command_end will handle those and provide better metadata
+        IF tg_tag NOT LIKE 'DROP %' THEN
```
eminano, line 33: This if check is the change introduced in this PR.

PR #729, opened 2026-02-17
fix: bound snapshot-store concurrency to prevent connection exhaustion

VikramChennai:
## Summary

- `markSnapshotInProgress` and `markSnapshotCompleted` used unbounded `errgroup` goroutines — one per schema. With N schemas (e.g. 41 for Zennagents), this opened N simultaneous connections from the snapshot-store pool just for bookkeeping INSERT/UPDATEs, exhausting the source database's connection slots (`SQLSTATE 53300`) before the data snapshot even started.
- Added `errgroup.SetLimit(4)` to both methods. The store operations are millisecond INSERT/UPDATEs — 4 concurrent workers process all schemas with negligible latency difference vs unbounded, while keeping pool usage constant regardless of schema count.
- Added concurrency test (`TestSnapshotRecorder_MarkInProgress_BoundedConcurrency`) that verifies peak concurrent store operations stay within the limit across 41 schemas.

## Root Cause

Instrumented all pgstream connection pools and ran against test Supabase source with `MaxConnections=3`. Observed:
- **snapshot-store**: N connections for N schemas (unbounded `errgroup.Go()`)
- **data-snapshot**: respected `MaxConnections` (exactly 3)
- **table-finder**: 1 connection

With Zennagents (41 schemas): 41 idle store connections + data snapshot connections + overhead = ~47+ total, crashing the 60-connection Supabase source.

## Test plan

- [x] `TestSnapshotRecorder_MarkInProgress_BoundedConcurrency` — 41 schemas, peak concurrent ops bounded to 4
- [x] All existing `TestSnapshotRecorder_*` tests pass unchanged
- [ ] Deploy to staging connector and verify no `53300` errors during snapshot

PR #731, opened 2026-02-19
[Snapshots] Set concurrency limit on snapshot recorder

eminano:
#### Description

This PR sets a limit to the concurrency of the snapshot recorder, preventing unbounded go routines to be started (one per schema), and potentially exhausting the connection pool resources from the source database. It uses the existing snapshot workers configuration which is already applied to the data snapshot. 

This can be a problem when there are many schemas being snapshotted in parallel.

##### Related Issue(s)

- Related to #729 

#### Type of Change

#### Testing

- [X] All existing tests pass


eminano:
> Is there a difference between `SnapshotWorkers` and `SchemaWorkers`? It is a bit confusing they are used interchangeably when setting the limit.

Yes, snapshot workers is concurrency per snapshot (number of concurrent schemas), while schema workers is how many workers per schema (number of concurrent tables). So in this case, I need to change the name for consistency. Thank you!

Issue #769, opened 2026-03-11
WAL streaming batch writer is slow for high-churn tables during catch-up


tsg:
## Problem

During pg-to-pg clone, after the snapshot completes, the WAL streaming catch-up phase replays all WAL events that accumulated during the snapshot window. For high-churn tables that generate a large volume of WAL events, the batch writer cannot keep up with the WAL backlog and the process appears "stuck".

During snapshot, we use the `BulkIngestWriter` which batches inserts efficiently via the COPY protocol. However, the WAL streaming phase uses the `BatchWriter` which executes each DML statement individually, with no batching or coalescing of operations.

PR #770, opened 2026-03-11
Add plan for coalescing operations during streaming

tsg:
Trying something new: this adds only the plan for solving #769, so we can discuss it before implementing.

Putting it in Draft mode because this is only meant for discussion.

### DIFF
--- a/batch_writer_coalesce_ops_plan.md
+++ b/batch_writer_coalesce_ops_plan.md
@@ -27,7 +27,8 @@ Additionally, the current architecture builds SQL strings *before* batching — 
 
 Batch raw WAL events instead of pre-built SQL strings, then build bulk SQL at execution time.
 
-- N DELETEs on the same table become: `DELETE FROM t WHERE "id" IN ($1, $2, ..., $N)`
+- N DELETEs on the same table (single-column PK) become: `DELETE FROM t WHERE "id" = ANY($1::bigint[])` — one array parameter
+- N DELETEs on the same table (composite PK) fall back to: `DELETE FROM t WHERE ("id","name") IN (($1,$2),($3,$4),...)` — individual parameters
 - N INSERTs on the same table become: `INSERT INTO t(cols) VALUES($1,$2),($3,$4),...`
 
 ### Architecture change
@@ -92,7 +93,9 @@ Change batch type from `batch.Batch[*query]` to `batch.Batch[*walMessage]`. New 
 1. Separate DDL and DML messages
 2. For DDL: build and execute queries via existing `ddlAdapter`
 3. For DML: walk messages in order, building "runs" of consecutive same-(schema, table, action) events:
-   - **DELETE run** → `DELETE FROM t WHERE "id" IN ($1,$2,...)` (single key) or `WHERE ("id","name") IN (($1,$2),($3,$4),...)` (composite key). IS NULL identity columns become a shared WHERE prefix.
+   - **DELETE run (single-column PK)** → `DELETE FROM t WHERE "id" = ANY($1::type[])` — all identity values collected into a single Go slice passed as one array parameter. The PG array type is derived from the identity column's `Type` field (see type mapping below).
+   - **DELETE run (composite PK)** → falls back to `DELETE FROM t WHERE ("id","name") IN (($1,$2),($3,$4),...)` with individual parameters, since `ANY` doesn't support tuple comparison. For composite PKs, respect the 65,535 parameter limit by splitting runs at ~60,000 params.
+   - IS NULL identity columns become a shared WHERE prefix for both variants.
    - **INSERT run** → `INSERT INTO t(cols) OVERRIDING SYSTEM VALUE VALUES($1,$2),($3,$4),...` with ON CONFLICT clause. Sequence setval queries generated separately.
    - **UPDATE** → individual UPDATE queries (no coalescing — each row has different SET values)
    - **TRUNCATE** → pass through as individual query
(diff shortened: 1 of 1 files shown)


batch_writer_coalesce_ops_plan.md
```diff
@@ -0,0 +1,164 @@
...
+| `pkg/wal/processor/postgres/postgres_wal_dml_adapter_bulk_test.go` | **New** — bulk builder tests |
+
+**Unchanged:**
+- `pkg/wal/processor/postgres/postgres_bulk_ingest_writer.go` — stays on `*query` with COPY
+- `pkg/wal/processor/batch/` — generic framework, no changes needed
+- `pkg/wal/wal_data.go` — no changes to WAL data model
+
+## Expected impact
+
+For the customer's workload (9,737 DELETEs in a single batch):
```
kvch, line 190: Did we see other columns in the `WHERE` condition in users' WAL events? In the future we could make it more flexible and batch DELETE statements with different column names. Example:
```sql
DELETE FROM t1 WHERE my_column in ($1, $2, $3)
```

batch_writer_coalesce_ops_plan.md
```diff
@@ -0,0 +1,164 @@
...
+
+Each `tx.Exec()` is a separate round-trip to the target database. With thousands of identical `DELETE FROM t WHERE "id" = $1` statements, the overhead of parsing, planning, and executing each statement individually dominates.
+
+Additionally, the current architecture builds SQL strings *before* batching — in `ProcessWALEvent`, each WAL event is immediately converted to an individual SQL string via `walEventToQueries()`, then the SQL string is batched. This means coalescing would require either re-parsing the SQL or double-transforming.
+
+## Proposed solution
+
+Batch raw WAL events instead of pre-built SQL strings, then build bulk SQL at execution time.
+
+- N DELETEs on the same table become: `DELETE FROM t WHERE "id" IN ($1, $2, ..., $N)`
```
kvch: Should we also consider composite primary keys?

kvch: Ah, i see that they are included.

batch_writer_coalesce_ops_plan.md
```diff
@@ -0,0 +1,164 @@
...
+
+### 4. Refactor BatchWriter.sendBatch — bulk query building
+
+**File:** `pkg/wal/processor/postgres/postgres_batch_writer.go`
+
+Change batch type from `batch.Batch[*query]` to `batch.Batch[*walMessage]`. New flow:
+
+1. Separate DDL and DML messages
+2. For DDL: build and execute queries via existing `ddlAdapter`
+3. For DML: walk messages in order, building "runs" of consecutive same-(schema, table, action) events:
```
kvch, line 95: Given that we already walk the messages in order, we can make the query aggregator logic a bit smarter, and consider adding interleaved DELETEs, if there is no INSERT or UPDATE statement that conflicts with them.

kvch (approved):
This is a good improvement. Hopefully, it will solve most of the issues for us. If not, there are several ways we can improve the process.

kvch (approved):
Solid plan, small comments

batch_writer_coalesce_ops_plan.md
```diff
@@ -0,0 +1,164 @@
...
+**File:** `pkg/wal/processor/postgres/postgres_batch_writer.go`
+
+Instead of calling `walEventToQueries()` and batching `*query`:
+
+1. Handle nil data / materialized views (skip)
+2. Handle DDL events: update schema observer, create `walMessage` with DDL data
+3. For DML: look up `schemaInfo` via schema observer (cached), create `walMessage` with raw `wal.Data` + `schemaInfo`
+4. Send `walMessage` to batch sender
+
+### 4. Refactor BatchWriter.sendBatch — bulk query building
```
kvch, line 87: What happens if the sending fails? Do we fall back to separate DELETE statements?

batch_writer_coalesce_ops_plan.md
tsg: Yes, we can add this later. In some situation we can prove that it's still correct.

batch_writer_coalesce_ops_plan.md
```diff
@@ -0,0 +1,164 @@
...
+DELETE FROM t WHERE id = 1    →  DELETE FROM t WHERE id = 1  (flush: action changes)
+INSERT INTO t (id) VALUES (1) →  INSERT INTO t (id) VALUES (1)
+DELETE FROM t WHERE id = 2    →  DELETE FROM t WHERE id = 2  (new run, only 1 event)
+```
+
+The ordering constraint is necessary for correctness — the INSERT between the two DELETEs may depend on the first DELETE completing (e.g., re-inserting a just-deleted row). Reordering across action boundaries could violate data integrity.
+
+In practice this works well for the target workload: WAL events from bulk operations on the source database (batch purges, accounting reconciliation, ETL loads) naturally produce long runs of the same operation on the same table, which coalesce effectively.
+5. Execute via existing `flushQueries` / `execQueries`
+6. Respect PostgreSQL's 65,535 parameter limit — split runs at ~60,000 params
```
kvch: I guess we can avoid this limit by using `WHERE id = ANY($1::bigint[])` because the array is just one parameter.

So the code would be:
```golang
fmt.Sprintf("DELETE FROM %s WHERE id = ANY($1::%s[])", table, idType)
args := []any{idArray}
```

This would perform better for multiple reasons:
* only one parameter is passed to postgresql, so the parameter binding overhead is smaller
* in the planning phase the original suggestion takes longer because the planner has to optimize a long-long condition with many `OR` operators. with the array, there is no such optimization

batch_writer_coalesce_ops_plan.md
tsg: WALs always refer to changes by their identify columns, so we don't have to support complex WHERE conditions.

batch_writer_coalesce_ops_plan.md
tsg: Maybe for simplicity for now we log as `DATALOSS`,  I think this is similar to how we do for batch inserts in the snapshot mode.

batch_writer_coalesce_ops_plan.md
tsg: great suggestion! Seems like we can do this when the types are well known (scalar types) and fallback to IN for the corner cases.

kvch:
Closing as the feature was implemented in #774

PR #774, opened 2026-03-12
Batch WAL operations in stream mode

### LINKED ISSUE #769: WAL streaming batch writer is slow for high-churn tables during catch-up

tsg:
#### Description

This is the implementation for the plan in #770 

Before this PR, we were batching write operations against the target in snapshot mode, but not in stream mode. This adds batching also in stream mode, as a performance optimisation.

##### Related Issue(s)

- Closes #769 
- Relates to #770 

#### Type of Change

- [ ] 📚 Documentation update
- [ ] 🔨 Build/CI changes
- [ ] 🧹 Code cleanup

#### Testing

- [ ] All existing tests pass

#### Checklist

- [ ] Self-review completed
- [ ] Code is well-commented

### DIFF
--- a/pkg/wal/processor/postgres/postgres_wal_dml_adapter_bulk.go
+++ b/pkg/wal/processor/postgres/postgres_wal_dml_adapter_bulk.go
@@ -41,7 +41,11 @@ func (a *dmlAdapter) getIdentityColumns(d *wal.Data) ([]wal.Column, error) {
 	case len(d.Identity) > 0:
 		return d.Identity, nil
 	case len(d.Metadata.InternalColIDs) > 0:
-		return a.extractPrimaryKeyColumns(d.Metadata.InternalColIDs, d.Columns), nil
+		cols := a.extractPrimaryKeyColumns(d.Metadata.InternalColIDs, d.Columns)
+		if len(cols) == 0 {
+			return nil, errUnableToBuildQuery
+		}
+		return cols, nil
 	default:
 		return nil, errUnableToBuildQuery
 	}


pkg/wal/processor/postgres/postgres_batch_writer.go
```diff
@@ -87,47 +92,123 @@ func (w *BatchWriter) Close() error {
...
+		}
+		return w.dmlAdapter.buildBulkDeleteQuery(events)
+	case "I":
+		events := make([]*wal.Data, len(run))
+		for i, m := range run {
+			events[i] = m.data
+		}
+		return w.dmlAdapter.buildBulkInsertQueries(events, run[0].schemaInfo), nil
+	default:
+		// UPDATE, TRUNCATE, and anything else: individual queries
```
tsg, line 203: This means we can still have the same type of issue with `UPDATE ... WHERE ...` statements that updates a lot of rows. Hopefully less common than the commit case.

pkg/wal/processor/postgres/postgres_wal_dml_adapter_bulk.go
```diff
@@ -0,0 +1,291 @@
...
+			args:        args,
+		})
+	}
+
+	// emit a single setval per sequence using the max value
+	for seqName, maxVal := range seqMaxValues {
+		queries = append(queries, &query{
+			table:  events[0].Table,
+			schema: events[0].Schema,
+			sql:    fmt.Sprintf("SELECT setval('%s', %d, true)", seqName, maxVal),
```
kvch: This should be parameterized because the sequence name might contain `'`.

pkg/wal/processor/postgres/postgres_batch_writer.go
```diff
@@ -87,47 +92,123 @@ func (w *BatchWriter) Close() error {
...
+							continue
+						}
+						return err
+					}
+				}
+				continue
 			}
+
+			// check if this message continues the current run
+			if len(currentRun) > 0 && (msg.data.Schema != runSchema || msg.data.Table != runTable || msg.data.Action != runAction) {
```
kvch, line 156: In case of materialized views `msg.data` is nil and `msg.isDDL` is false. So `msg.data.Schema` is nil and this check can panic. We could add a guard that checks if the message is empty, and skip those empty messages.

pkg/wal/processor/postgres/postgres_wal_dml_adapter_bulk.go
tsg: Ha TIL sequence names can contain '. Good catch, thanks!

pkg/wal/processor/postgres/postgres_batch_writer.go
tsg: Added a guard against empty message.

PR #787, opened 2026-04-02
Plan for simpler upgrading procedure

tsg:
This contains only the plan for simplifying the upgrading procedure from 0.9.x to 1.0.

The idea is that instead of having to clean up with pgstream 0.9 and then re-init with 1.0, an `--upgrade` flag on 1.0 handles both the clean up and the new init.

Seems the clean-up is easy/safe enough that we can do it directly.


kvch (approved):

kvch:
The feature was implemented in #788

PR #790, opened 2026-04-08
feat: support environment variables in YAML configuration

### LINKED ISSUE #383: [Config] Support for env variables in yaml configuration
Add support to provide environment variables in the yaml configuration. This would allow to avoid exposing secrets in the configuration. 

The yaml golang library we're using currently doesn't support it (see https://github.com/go-yaml/yaml/issues/437), but should be easy to add leveraging the [`os.ExpandEnv(str)`](https://pkg.go.dev/os#ExpandEnv) functionality.

mvanhorn:
## Summary

Adds `os.ExpandEnv` support to YAML config parsing so users can reference environment variables like `${DB_PASSWORD}` in config files instead of hardcoding secrets.

Applied to all three YAML-backed parsing paths:
- `LoadFile` (viper): reads file manually, expands env vars, feeds to `viper.ReadConfig`
- `ParseStreamConfig` (yaml.Unmarshal): expands before the case-sensitive YAML parse
- `ParseTransformerConfig`: expands before transformation rule parsing

Also handles the transformer rules merge path (`PGSTREAM_TRANSFORMER_RULES_FILE`).

Example config:
```yaml
listener:
  postgres:
    conn: "postgres://${PG_USER}:${PG_PASSWORD}@${PG_HOST}:5432/mydb"
```

Fixes #383

This contribution was developed with AI assistance (Codex).


kvch (changes requested):
This is a great feature! However, the implementation breaks the transformer configurations. Could you please fix the code so it only expands `${...}` variables?

mvanhorn:
Fixed - expansion now only touches ${...} forms (valid env identifier inside braces). Bare $VAR and transformer template/regex dollar syntax ($1, ${1}, $$) pass through untouched; added table tests covering those cases plus a ParseTransformerConfig test asserting transformer configs survive expansion.

kvch (approved):

kvch:
Thank you for the fixes!

mvanhorn:
Appreciate the merge @kvch, glad env vars in YAML config made it in.

mvanhorn:
Thanks for the quick review and merge, @kvch. Glad env var support in the YAML config made the cut.

PR #805, opened 2026-05-09
[bugfix] Support composite primary keys in WAL metadata

blurskye:
## Bug

`on_conflict_action: update` can generate invalid SQL for tables with composite primary keys.

The Postgres writer already supports multiple identity columns when it builds `WHERE` and `ON CONFLICT` clauses. The issue is earlier in the pipeline: the WAL injector only stored the first primary-key column in `Metadata.InternalColIDs`.

A small table like this is enough to trigger it:

```sql
CREATE TABLE sensor_samples (
  sensor_id text NOT NULL,
  sampled_at timestamptz NOT NULL,
  channel text NOT NULL,
  reading double precision NOT NULL,
  PRIMARY KEY (sensor_id, sampled_at, channel)
);
```

Before this patch, pgstream could build:

```sql
ON CONFLICT ("sensor_id") DO UPDATE SET ...
```

Postgres rejects that because `sensor_id` is not unique by itself. The valid conflict target is the full primary key:

```sql
ON CONFLICT ("sensor_id", "sampled_at", "channel") DO UPDATE SET ...
```

## Why it happens

The metadata flow looked like this:

```mermaid
graph LR
    A[Postgres table metadata] --> B[WAL injector]
    B --> C[Metadata.InternalColIDs]
    C --> D[Postgres writer]
    D --> E[WHERE and ON CONFLICT SQL]
```

The writer was ready for composite keys, but the injector only gave it one key column. So the writer did the best it could with incomplete metadata and produced a conflict target that did not match any unique constraint.

## Fix

This changes the injector to put every primary-key column into `Metadata.InternalColIDs`.

For tables with a primary key:

```text
before: [sensor_id]
after:  [sensor_id, sampled_at, channel]
```

Tables without a primary key keep the existing fallback behavior. pgstream still uses the first not-null unique column as the row identity.

## Extra cleanup

This also makes duplicate table-mapping creation safe:

```sql
INSERT INTO pgstream.table_ids (oid) VALUES (table_oid)
ON CONFLICT (oid) DO UPDATE SET oid = EXCLUDED.oid
RETURNING id;
```

That prevents concurrent metadata lookups from racing when they try to create the same table mapping.

## Tests

Ran:

```sh
go test ./pkg/wal/processor/injector ./pkg/wal/processor/postgres ./internal/migrator ./pkg/stream
```

I also smoke-tested a Postgres 16 source and target with:

- initial snapshot
- live inserts, updates, and deletes
- `on_conflict_action: update`
- a table with a three-column primary key
- pgstream restart with pending WAL

The target caught up and matched the source after replication resumed.


kvch (approved):

kvch:
Thank you for the contribution!

PR #809, opened 2026-05-11
fix(postgres): register pgvector types

rieg-ec:
#### Description

came across this issue when using pgstream with a database that uses `vector` extension with a custom size `#  embedding   :vector(3072)`

i patched the pgstream pg_utils.go  to call `pgxvec.RegisterTypes` and it worked. Without this, pgx does not know how to encode/decode pgvector values and Postgres targets can fail when a replicated table contains `vector` columns.

not sure if this is the right place to do it, it was highly vibecoded, but it seemed to work for us so i'll let this PR in case it's useful

##### Related Issue(s)

- Related to pgvector extension support

#### Type of Change

- [ ] 📚 Documentation update
- [ ] 🔨 Build/CI changes
- [ ] 🧹 Code cleanup

#### Changes Made

- Register pgvector types through `github.com/pgvector/pgvector-go/pgx` in the shared Postgres connection type setup.
- Ignore the expected `vector type not found` registration error when the extension is not installed.
- Add unit coverage for the pgvector registration error handling.

#### Testing

- [x] All existing tests pass

Commands run:

```bash
go test ./internal/postgres
go test ./...
```

#### Checklist

- [x] Self-review completed
- [x] Code is well-commented

#### Additional Notes

-


internal/postgres/pg_utils.go
```diff
@@ -161,9 +164,23 @@ func registerTypesToConnMap(ctx context.Context, conn *pgx.Conn) error {
 		})
 	}
 
+	if err := handlePgvectorRegisterTypesErr(pgxvec.RegisterTypes(ctx, conn)); err != nil {
```
kvch: Could you please refactor these lines to be similar to what we do for the extension `hstore`?

Something along the lines of:
```golang
 var vectorOID uint32
  if err := conn.QueryRow(ctx, "SELECT to_regtype('vector')::oid").Scan(&vectorOID); err == nil && vectorOID != 0 {
      if err := pgxvec.RegisterTypes(ctx, conn); err != nil {
          return fmt.Errorf("registering pgvector types: %w", err)
      }
  }
```

kvch (commented):
Thank you for the contribution. Asking for a small refactoring.

kvch (approved):

PR #823, opened 2026-05-21
Add support for human readable byte size settings

kvch:
#### Description

Config values that set bytes (`_BYTES` suffix) both in env variables and YAML configuration accept human readable suffixes. In the past, the documentation said that we support it. However, we did not. So `pgstream` silently used the defaults instead of the configured values.

From now on, `pgstream` parses values like `15MiB`.

Another behaviour change is that now the value cannot be parsed, `pgstream` fails to start. Previously, in case of errors, `pgstream` started with the defaults.

Unset values continue to work the same as before.

#### Type of Change

- [ ] 📚 Documentation update
- [ ] 🔨 Build/CI changes
- [ ] 🧹 Code cleanup

#### Changes Made

- Indirect dependencies became direct dependencies
- Stricter configuration check

#### Testing

- [x] All existing tests pass

#### Checklist

- [x] Self-review completed
- [x] Code is well-commented

### DIFF
--- a/cmd/config/bytesize.go
+++ b/cmd/config/bytesize.go
@@ -1,0 +1,72 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package config
+
+import (
+	"fmt"
+	"strings"
+
+	units "github.com/docker/go-units"
+	mapstructure "github.com/go-viper/mapstructure/v2"
+	"github.com/spf13/viper"
+	"gopkg.in/yaml.v3"
+)
+
+// byteSize represents a size in bytes that can be configured either as a plain
+// integer or as a human-readable string with a unit suffix (e.g. "64MiB" or
+// "1GiB"). Units are case-insensitive and interpreted as binary multiples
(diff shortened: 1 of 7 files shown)


cmd/config/bytesize.go
```diff
@@ -0,0 +1,72 @@
...
+		return 0, fmt.Errorf("%s: %w", key, err)
+	}
+	return n, nil
+}
+
+// byteSizeDecodeHook builds the mapstructure hook used to decode the YAML
+// config. viper.DecodeHook replaces viper's default hook rather than appending
+// to it, so the duration and slice hooks are re-included here alongside the
+// text unmarshaller hook that decodes byteSize fields.
+func byteSizeDecodeHook() mapstructure.DecodeHookFunc {
```
kvch, line 45: At the moment we do not have options that contain string slices or duration. So we could skip it in theory. But I am keeping these parsers to make sure if someone adds a new option, we can parse it correctly.

exekias (approved):

PR #831, opened 2026-06-02
Add support for cube data type

### LINKED ISSUE #801: cube dimension is too large
**Describe the bug**
Pgstream fails to copy cube columns with the following error:
```
2026-04-28T14:57:37.586795+01:00 ERR logger.go:51 > failed to send batch
  error.message="copy from: program limit exceeded: cube dimension is too large"
  module=postgres_bulk_ingest_writer
  schema=musicbrainz
  table=medium_index
```
This error doesn't appear when using pg_dump and pg_restore.

**To Reproduce**
Use `xata clone` with the attached sql.

[cube_repro.sql](https://github.com/user-attachments/files/27235936/cube_repro.sql)

**Setup (please complete the following information):**
```
┌───────┬────────┬──────────┐
│ xata  │ pgroll │ pgstream │
├───────┼────────┼──────────┤
│ 1.2.0 │ 0.16.1 │ 0.9.11   │
└───────┴────────┴──────────┘
(PostgreSQL) 18.3
```

kvch:
#### Description

This PR adds support for [cube](https://www.postgresql.org/docs/current/cube.html) data type. The data type is used for representing multidimensional cubes.

At the moment snapshotting a table with a `cube` column fails on the COPY step:

```
copy from: program limit exceeded: cube dimension is too large
  module=postgres_bulk_ingest_writer
```

The root cause of the issue `pgx`'s `CopyFrom` is hardcoded to `COPY ... FROM STDIN BINARY` and emits each value through the codec registered for its OID. `cube` has no binary codec in pgx; the fallback writes the raw text bytes into a binary frame, and on the receiving side Postgres parses them through the cube binary input function. The first 4 bytes of `(1,2,3,4,5)` are read as a dimension count, hence "cube dimension is too large".

`pg_dump`/`pg_restore` are unaffected because they use `COPY ... FROM STDIN` (text format), which routes each value through the per-type text input function (`cube_in`).

##### Related Issue(s)

- Fixes #801 

#### Type of Change

- [ ] 📚 Documentation update
- [ ] 🔨 Build/CI changes
- [ ] 🧹 Code cleanup

#### Changes Made

- **`internal/postgres/pg_tx.go`**: keep `Txn.CopyFrom` on pgx's fast binary path and add a new `Txn.CopyFromText` that runs the postgres COPY protocol in text format via `pgconn.PgConn.CopyFrom(... "COPY … FROM STDIN")`. Values are encoded with `TypeMap.Encode(oid, TextFormatCode, …)` and the stream uses COPY text escaping (`\t \n \r \\ \b \f \v`). The whole COPY payload is serialised into a single `bytes.Buffer` and handed to `pgconn.CopyFrom` via `bytes.Reader`; unescaped runs of bytes are emitted in one `Write` call to keep the hot path cheap.

- **bulk ingest writer dispatch**: `buildInsertQueries` / `buildBulkInsertQueries` compute a `needsTextCopy` flag on the resulting `*query` by checking each column's postgres type against a small `textOnlyCopyTypes` set (just `cube` here). `BulkIngestWriter.copyFromInsertQueries` reads that flag and dispatches to `tx.CopyFromText` only for affected batches; everything else stays on binary COPY, so non-cube snapshots keep their original throughput.

- **`internal/postgres/pg_utils.go`**: generalise the per-type if-blocks into a table-driven `extensionTypes` slice. Each entry binds an extension to a register function. `hstore` and `vector` keep their existing codec/library wiring; `cube` is registered with `pgtype.TextCodec` so the source-side read path (`rows.Values()`) returns cube values as strings rather than raw `[]byte`, which is what the text-format encoder expects.

#### Testing

- [x] All existing tests pass

#### Checklist

- [x] Self-review completed
- [x] Code is well-commented

#### Additional Notes

Text-format COPY is slower than binary on numeric/timestamp/uuid columns (bigger wire payload + per-type input functions instead of memcpy on the destination).

**The conditional dispatch confines the cost to tables that actually contain a cube column. Every other snapshot path keeps the original binary throughput.**

Measured locally with `BenchmarkCopyFrom` in `internal/postgres/pg_tx_bench_test.go` (Postgres 17 in testcontainers, `-benchtime=20x`):

| Shape | n | Binary | Text | Slowdown |
| --- | --- | --- | --- | --- |
| numeric-heavy (int8, timestamptz, uuid) | 10,000 | 42.05 ms / 14.27 MB/s | 44.51 ms / 13.48 MB/s | 6% |
| text-heavy (text, jsonb) | 10,000 | 55.25 ms / 27.33 MB/s | 63.04 ms / 23.95 MB/s | 14% |

Reproduce: `PGSTREAM_INTEGRATION_TESTS=true go test -bench=BenchmarkCopyFrom -benchtime=20x -run=^$ ./internal/postgres/`.

### DIFF
--- a/internal/postgres/pg_tx_bench_test.go
+++ b/internal/postgres/pg_tx_bench_test.go
@@ -104,7 +104,6 @@ func BenchmarkCopyFrom(b *testing.B) {
 	}
 
 	for _, shape := range shapes {
-		shape := shape
 		// reset table once per shape so benchmarks share state
 		_, err := pool.Exec(ctx, "DROP TABLE IF EXISTS bench_numeric, bench_text")
 		require.NoError(b, err)
--- a/internal/postgres/pg_tx_bench_test.go
+++ b/internal/postgres/pg_tx_bench_test.go
@@ -1,0 +1,178 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package postgres_test
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+	pglib "github.com/xataio/pgstream/internal/postgres"
+	"github.com/xataio/pgstream/internal/testcontainers"
+)
+
+// BenchmarkCopyFrom compares pgx's binary-format COPY (Txn.CopyFrom) against
+// the text-format implementation (Txn.CopyFromText) on the same row payload.
+//
+// Two table shapes:
+//   - numeric: int8/timestamp/uuid columns — binary's strongest case (smaller
+//     wire payload, no per-type input function on the destination).
+//   - text:    text/varchar/jsonb columns — the gap shrinks because both
+//     modes ultimately route through text-like decoding.
+//
+// Each benchmark runs one COPY of `rows` rows per iteration. The container
+// is spun up once for the whole run and reused.
+//
+// Gated on PGSTREAM_INTEGRATION_TESTS so `go test ./...` doesn't pay the
+// container cost.
+//
+// Usage:
+//
+//	PGSTREAM_INTEGRATION_TESTS=true go test -bench=BenchmarkCopyFrom \
+//	    -benchtime=10x -run=^$ ./internal/postgres/
+func BenchmarkCopyFrom(b *testing.B) {
+	if os.Getenv("PGSTREAM_INTEGRATION_TESTS") == "" {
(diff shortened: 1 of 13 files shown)


internal/postgres/pg_tx_bench_test.go
```diff
@@ -0,0 +1,177 @@
+// SPDX-License-Identifier: Apache-2.0
```
kvch, line 1: This file can be removed. This is the test I used for benchmarking the changes to data copying. I have checked it in if anyone wants to run the benchmarks themselves they can do it.

exekias (approved):

PR #896, opened 2026-06-12
Add schema compatibility validation for pgstream

kv1sidisi:
- Add a `pgstream validate schema` subcommand for PostgreSQL preflight checks.
- Check source table type resolution and replica identity safety before replication starts.
- Document the new command and cover the checker with focused unit tests.


kvch:
Thank you for the contribution! The coverage failure is expected. However, the code generator failure is real.

Also, this issue is part of a greater effort in making `pgstream` runs more predictable. We are happy to take the contribution. But we might change the interface of this command. Maybe introduce a new command altogether named `check` that would include even more checks. There is no need for you to apply this change yet, as we are still working on the details of the new functionality.

kvch (approved):

PR #946, opened 2026-06-26
fix: reload log level on SIGHUP

### LINKED ISSUE #943: Reload log level on `SIGHUP`
## What to build

On `SIGHUP`, re-read the configured log level and apply it to the running process live instead of treating `SIGHUP` as a shutdown signal.

Today `SIGHUP` is registered alongside `SIGINT`/`SIGTERM` in `cmd/root_cmd.go` and cancels the root context (shutdown).

## Requirements:
  - [x] `SIGHUP` re-reads logging.level from the config file and applies it to the running logger with no restart and no shutdown
  - [x] An invalid level is rejected and the previous level keeps running (never crash the process)
  - [x] The level change is logged at a level that is always visible
  - [x] Existing `SIGINT`/`SIGTERM` shutdown behavior is unchanged
  - [x] `SIGHUP` with no change in level is a no-op

Saithej2k:
Summary
- Handle SIGHUP separately from shutdown signals.
- Re-read the configured log level and apply valid changes to the running logger.
- Keep the existing level when reload fails or the new level is invalid.

Testing
- `go test ./cmd ./internal/log/zerolog`

Closes #943

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -25,14 +25,25 @@ const (
 )
 
 func Load() error {
-	return LoadFile(viper.GetString("config"))
+	return loadFile(viper.GetString("config"), true)
+}
+
+// Reload re-reads the configured file without writing startup messages.
+func Reload() error {
+	return loadFile(viper.GetString("config"), false)
 }
 
 func LoadFile(file string) error {
+	return loadFile(file, true)
+}
+
+func loadFile(file string, announce bool) error {
 	if file == "" {
 		return nil
 	}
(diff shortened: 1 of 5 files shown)


kvch:
Thank you for the contribution. Two PRs were submitted to implement log level reload. I selected the [other PR](https://github.com/xataio/pgstream/pull/945) because it uses the atomic `SetGlobalLevel` function to change the log level. This implementation leads to a data race because it mutates the `Logger` instance directly.

We welcome your future contributions.

Saithej2k:
Thanks for the explanation. That makes sense; I missed the race from mutating the logger directly. I’ll keep the atomic SetGlobalLevel approach in mind for future changes.

Issue #897, opened 2026-06-17
Add `pgstream check` command for pre-migration sanity checks


kvch:
## Problem

A pgstream migration can fail many hours in:

- a `REPLICA IDENTITY` misconfiguration that only shows up the first time an UPDATE flows
- an extension missing on the target that pg_restore only notices at COPY time
- the replication role missing the `REPLICATION` attribute
- `max_replication_slots` already exhausted
- a target Postgres version that can't accept some syntax the source uses
- a column type pgstream can't decode buried inside an otherwise-clean schema
- `max_connections` too low for the configured worker count

Users who've already spent 10+ hours waiting for a snapshot or replication catch-up shouldn't be discovering setup issues at hour 10 that could have been flagged in 30 seconds before anything ran.

## Requirements

- A single CLI invocation runs every applicable check using the same configuration source as `run` or `snapshot`.
- Each finding has severity (error / warning / info) and an actionable message.
- Inapplicable checks (e.g. CREATEROLE when `roles_snapshot_mode: disabled`) are skipped with a one-line reason.
- Exit codes: `0` clean, `1` error-tier findings, `2` validator failure. `--strict` upgrades warnings to errors.
- Output supports human-readable and `--json` formats.

## Solution

Add a pre-migration check (working name: `pgstream check`) that runs every check that can be answered statically (connectivity, privileges, version, replication setup, schema compatibility, resource capacity) and returns a structured report.

The command reads the same YAML/env/flag configuration that `pgstream run` and `pgstream snapshot` consume, so users can validate the exact config they're about to migrate with.

Exit codes mirror `validate rules`: `0` clean, `1` blocking findings, `2` validator itself failed. `--json` produces a machine-readable report; default is human-readable with severity-tagged sections.

### Connectivity & identity @kvch 
- [x] Source is reachable
- [x] Target is reachable
- [x] Source and target are not the same host/database (catches accidental same-DB config)
- [x] Source and target Postgres versions are compatible (target >= source; flag known cross-version pitfalls)

### Replication setup @kvch 
- [x] `wal_level = logical` on source
- [x] `wal2json` is installed and loadable on source
- [x] At least one replication slot is available (`pg_replication_slots` vs `max_replication_slots`)
- [x] Source role has the `REPLICATION` attribute
- [x] Per-table `REPLICA IDENTITY` is sufficient for UPDATE/DELETE replication (default + PK, or FULL, or explicit USING INDEX)
[...]

PR #947, opened 2026-06-28
  Add preflight check for source sequence SELECT privileges

lghuy05:
#### Description

Add a second access preflight check that verifies the source role can SELECT from in-scope sequences used by snapshot tables.

This extends pgstream -check --access beyond table read privileges so sequence-value snapshot failures are caught earlier with a direct remediation message. 

##### Related Issue(s)

- Related to #897 

#### Type of Change

- [ ] 📚 Documentation update
- [ ] 🔨 Build/CI changes
- [ ] 🧹 Code cleanup

#### Changes Made

- add `SourceSequenceSelectPrivilegesCheck` to `pkg/stream/preflight/access.go`
- query table-to-sequence-relationships from Postgres catalogs and evaluate access with has_sequence_privilege(..., 'SELECT')
- update `BuildAccessChecks` to run both table and sequence access checks with the same shared source connection

#### Testing

- [x] All existing tests pass

#### Checklist

- [x] Self-review completed
- [ ] Code is well-commented

#### Additional Notes

- I didn't fully verify the integration job


kvch:
Could you please rebase your branch?

pkg/stream/preflight/access.go
```diff
@@ -35,6 +46,31 @@ WHERE c.relkind IN ('r', 'p')
...
+FROM pg_class t
+JOIN pg_namespace tn ON tn.oid = t.relnamespace
+JOIN pg_attribute a ON a.attrelid = t.oid
+JOIN pg_attrdef ad ON ad.adrelid = t.oid AND ad.adnum = a.attnum
+JOIN pg_depend d ON d.refobjid = t.oid AND d.refobjsubid = a.attnum
+JOIN pg_class s ON s.oid = d.objid
+JOIN pg_namespace sn ON sn.oid = s.relnamespace
+WHERE t.relkind IN ('r', 'p')
+  AND s.relkind = 'S'
+  AND d.deptype = 'a'
```
kvch: Does this condition include generated [identity columns](https://www.postgresql.org/docs/current/ddl-identity-columns.html)? Could you please add a test to verify that those columns are checked as well?

lghuy05: Thanks, I missed the identity-column case. I've updated the sequence privilege query to account for identity-backed sequences and added integration test covering `GENERATED ALWAYS AS IDENTITY`

pkg/stream/preflight/access.go
```diff
@@ -35,6 +46,30 @@ WHERE c.relkind IN ('r', 'p')
 ORDER BY n.nspname, c.relname
 `
 
+const sourceSequenceSelectPrivilegesQuery = `
+SELECT
```
kvch, line 50: I've been testing this a bit more, and found that this query cannot find all sequences that are snapshotted. It is missing the sequences that belong to the schema, but not owned by any columns. Sorry for pointing it out now. Could you please make sure the check includes:
* sequences of the table
* generated identity columns of the table
* all sequences in the selected schema

lghuy05: Thanks. I've updated the check to cover `column-owned sequences`, `identity-backed sequences` and `free-standing sequences` in selected schemas and added test for those cases.

kvch (approved):
Thank you for both the contribution and your patience.

PR #976, opened 2026-07-07
Fail snapshots on dropped Postgres writes if in strict mode

### LINKED ISSUE #973: [Snapshots] Fail snapshots when Postgres writes are dropped
## What to build

Postgres snapshots should fail if the target Postgres writer drops a non-internal DATALOSS query, including failures that happen while draining the final batch during snapshot shutdown.

Current `main` can log a dropped query and still return success from `stream.Snapshot`, which makes an incomplete target look successfully snapshotted.

## Requirements
- [ ] Snapshot mode uses the Postgres writer strict mode by default
- [ ] Continuous replication keeps the existing default unless `strict_mode` is configured
- [ ] Batch sender `Close` returns recorded send/drain failures and processors propagate them
- [ ] `PGSTREAM_POSTGRES_WRITER_STRICT_MODE` is bound from env and documented
- [ ] Regression coverage uses a synthetic schema only

## Context

#959 added Postgres writer strict mode and dropped-query metrics. The remaining gap is making snapshots fail loudly by default, and returning final-batch errors from `Close` so snapshot callers can observe them.

danddanddand:
## Summary
- add an opt-in `strict_mode` option to the Postgres writer (plain `bool`, defaults to `false` for both replication and snapshots) so dropped writes can surface as errors instead of being silently logged
- return recorded batch send/drain failures from `Sender.Close` and propagate them through Kafka, search, Postgres batch, and Postgres bulk-ingest writer close paths
- bind and document `PGSTREAM_POSTGRES_WRITER_STRICT_MODE`
- add a synthetic inherited child-table snapshot regression (with `strict_mode` enabled) that fails when dropped rows are reported as success
- narrow reused snapshot integration subtests from broad `*.*` table selection to the tables each subtest creates, so strict-mode snapshots do not trip over rows left behind by earlier subtests (same hunks as #978; they merge cleanly in either order)

## Why these are together

The snapshot regression depends on both pieces: strict mode turns the duplicate-key write into an error, and `Close` propagation returns final-batch failures that happen while the snapshot processor drains at shutdown. Without the close-error propagation, the snapshot can still report success after the writer logs a dropped query.

## Testing
- `make test`
- `make integration-test`

Fixes #973


kvch (changes requested):
This is also a great fix, thank you. But I would like to preserve the behaviour of snapshotting to avoid surprises for existing users. Please change the type of `StrictMode` to `bool` and keep the default value as well.

danddanddand:
Reworked per review: `StrictMode` is back to a plain `bool` and the default stays `false` for both replication and snapshots — snapshots no longer force-enable it, so existing snapshot behaviour is unchanged and strict mode is purely opt-in. Docs updated to match, and the row-loss regression test now enables `strict_mode` explicitly.

kvch (approved):

kvch:
Thank you!

PR #977, opened 2026-07-07
Skip pgstream metadata DML for Postgres targets

### LINKED ISSUE #974: [Postgres] Skip internal pgstream metadata DML on Postgres targets
## What to build

The Postgres target adapter should ignore non-DDL rows for the internal `pgstream` schema because those rows describe source-side metadata and should not be materialized into a plain Postgres target.

## Requirements
- [ ] Non-DDL events for `pgstream.*` become keepalive/no-op events for Postgres targets
- [ ] DDL events are still processed normally
- [ ] Unit coverage proves internal metadata DML is skipped

danddanddand:
## Summary
- filter non-DDL WAL rows from the internal `pgstream` schema in the Postgres WAL listener before target processing
- convert skipped internal metadata rows into nil-data keepalive/checkpoint events so downstream processors can still advance commit positions
- preserve normal DDL logical message handling, including DDL events that carry the `pgstream` schema name
- add listener coverage for skipped internal metadata DML and processed `pgstream` schema DDL

## Testing
- `go test ./pkg/wal/listener/postgres ./pkg/wal/processor/postgres`
- `make test`
- `make integration-test`

Fixes #974


kvch (changes requested):
This is indeed an issue in `pgstream` at the moment. So thank you for addressing it. However, this is not the right layer to drop the events from `pgstream` schema updated. This issue impacts all of the targets of `pgstream` (Kafka, webhooks, OpenSearch). Your PR only addresses it in pg->pg replication. Furthermore, we could filter out these events earlier in the pipeline.

Could you please rework the PR so the filtering happens during the WAL event processing (`processWALEvent` in `wal_pg_listener.go`)?

Until this fix is released, you can filter out `pgstream` schema using the `filter_tables` plugin setting:

```yaml
source:
    postgres:
      url: "postgres://user:pass@source:5432/mydb"
      replication:
        plugin:
          filter_tables: "pgstream.*"
```

danddanddand:
Moved the filtering into `processWALEvent` in the WAL listener as requested — internal `pgstream` DML now becomes a keep-alive (commit position still checkpointed) before reaching any target, and the writer-level filter is removed. On the red integration job: the failure is the shared replication slot contention addressed in #978 (`replication slot "pgstream_postgres_slot" is active for PID …` in the log); the full integration suite passes locally on this branch head.

kvch (approved):

PR #978, opened 2026-07-07
Isolate Postgres integration replication state

### LINKED ISSUE #975: [Tests] Isolate Postgres integration replication state
## What to build

Postgres integration helpers should avoid stale WAL and stale snapshot rows leaking between tests.

## Requirements
- [ ] Listener integration tests use per-test replication slots and clean them up
- [ ] Reused snapshot subtests snapshot only the current tables instead of broad `*.*`
- [ ] Async run/snapshot helper errors are surfaced during cleanup
- [ ] Existing integration target still passes

danddanddand:
## Summary
- use per-test Postgres replication slots for integration listeners and clean them up after stream shutdown
- run shared integration setup in migrations-only mode so it no longer creates an unused default replication slot
- surface async `runStream`/`runSnapshot` errors from test cleanup
- narrow reused snapshot subtests from broad `*.*` table selection to the tables created by each subtest

## Testing
- `git diff --check`
- `make test`
- `make integration-test`

Fixes #975

> Note: the snapshot subtest table-narrowing hunks in `snapshot_pg_integration_test.go` are intentionally duplicated in #976 (they are needed there for strict-mode snapshots to pass CI). The hunks are identical, so the two PRs merge cleanly in either order.


kvch (approved):
Thank you! If you are up for it, could you please remove the default slot created in `setup_test.go`? It is not used anymore.

danddanddand:
Done — `setup_test.go` now runs `Init` with `MigrationsOnly: true`, so the pgstream migrations still run but the unused default replication slot is no longer created.

PR #979, opened 2026-07-08
Extract ctid strategy

### LINKED ISSUE #927: [Snapshots] PK snapshot: tableReader seam refactor
## What to build
Extract today's ctid logic in `pg_snapshot_generator.go` behind a `tableReader` interface (`beginSchema`/`readTable`); `ctidReader` = current behavior verbatim.

## Requirements
- [ ] No behavior change; existing unit + integration tests green
- [ ] `beginSchema` exports the transaction snapshot only when the schema has ≥1 ctid table

kvch:
#### Description

Refactors `pkg/snapshot/generator/postgres/data` so the `tableReader` strategy is a factory of read sessions instead of a single object the generator reaches into.

Now a `tableReader` only opens a session per schema, and the session (already bound to its schema and strategy-specific state) does the reading. `ctidReader` produces a `ctidSession` that owns the `snapshotID`, removing it from the signatures of `readTable`, `getTableInfo`, and `snapshotTableRange`, making it impossible to pair a session with the wrong reader, instead of relying on documentation.

  ##### Related Issue(s)

  - Closes #927 

  #### Type of Change

  - [x] 🔧 Refactoring (no functional changes)

  #### Changes Made

  - Introduce `tableReader`/`readSession` interfaces (`table_reader.go`) and extract the ctid strategy into `ctidSession`/`ctidReader` (`ctid_table_reader.go`), which owns the `snapshotID` and `totalBytes` measurement
  - Move `getSnapshotSchemaTotalBytes` onto `ctidSession.totalBytes` — the generator owns the progress bar, the strategy owns how the bytes behind it are measured
  - Wrap the session as well as the reader in `instrumentedTableReader`; `beginSchema` now opens a span too, so table spans get a schema parent
  - Extract `rowSink` (`row_sink.go`): the row → WAL event → processor → progress path is shared regardless of which rows a strategy reads, so it no longer sits inlined in `snapshotTableRange`
  - Extract `snapshotTx` (`snapshot_tx.go`) and `progressTracker` (`progress_tracker.go`) to collapse the duplicated `progressTracking` bool + `progressBars` map into a single value with `set`/`advance`/`complete` methods; a zero-value
  tracker no-ops, removing per-call guards
  - Build the reader in `newTableReader`, so choosing a strategy and applying decorators happens in one place instead of in `NewSnapshotGenerator`

  #### Testing

  - [x] All existing tests pass

  #### Checklist

  - [x] Self-review completed
  - [ ] Code is well-commented


pkg/snapshot/generator/postgres/data/instrumented_table_reader.go
```diff
@@ -0,0 +1,63 @@
...
+	})
+}
+
+type instrumentedReadSession struct {
+	tracer  trace.Tracer
+	session readSession
+}
+
+func (i *instrumentedReadSession) readTable(ctx context.Context, table *table) (err error) {
+	ctx, span := otel.StartSpan(ctx, i.tracer, "tableReader.ReadTable", trace.WithAttributes([]attribute.KeyValue{
```
tsg, line 47: Claude flagged this was called `tableSnapshotGenerator.SnapshotTable` and was used in one of the pgstream dashboards: build/docker/signoz/dashboards/pgstream.json

tsg (approved):
Minor comment left.

PR #995, opened 2026-07-10
Parallelize COPY workers

kvch:
#### Description

During a pg2pg snapshot, each table's rows are written to the target through a single COPY stream. Once the smaller tables finish, the snapshot spends most of its wall-clock waiting on the largest table draining through that one stream.  Snapshot reads are already parallel (`snapshot.data.table_workers` splits each table into page ranges), but everything funneled into one writer per table, so one big table dictated total snapshot time.

This PR adds a new option to control how many COPY streams write each table to the target in parallel:

```
target:
  postgres:
    url: "postgresql://..."
    bulk_ingest:
      enabled: true
      copy_workers: 8   # default 8; set to 1 for the previous behavior
```

or via environment variable:

`PGSTREAM_POSTGRES_WRITER_BULK_INGEST_COPY_WORKERS=8`

It only applies to the snapshot bulk-ingest path (`bulk_ingest.enabled`: true). Replication is untouched.

##### How it works

Rows read from the source are grouped into batches as before; completed batches are handed to a pool of `copy_workers` senders that each run their own `COPY` into the target table concurrently. This is safe on the snapshot path because bulk ingest handles independent inserts only, indexes and constraints are restored after the data load, and batches carry no ordering requirement. A global cap keeps the total number of concurrent `COPY`s across all tables within the target connection pool, so many small tables can't starve the pool. Batch sizing is adjusted automatically so the memory footprint stays the same as before (one large batch is replaced by `copy_workers` smaller ones). If a batch fails, the first error wins, in-flight COPYs are cancelled, and the usual whole-table retry semantics apply unchanged.

##### Benchmarks

Benchmarks showed 1.7x speedup on smaller tables (it was limited my laptop's bandwidth). I am still running benchmarks for big tables.

Benchmarks with 15GB table:

| workload | `copy_workers=1` | `copy_workers=8` | speedup |
|---|---|---|---|
| 15GB heap / 55.5M rows, data phase | 32.8 min (7.8 MB/s) | 8.9 min (28.7 MB/s) | 3.67x |
| 15GB heap / 55.5M rows, end-to-end | 35m59s | 12m00s | 3.00x |

#### Type of Change

- [ ] 📚 Documentation update
- [ ] 🔨 Build/CI changes
- [ ] 🧹 Code cleanup

#### Testing

- [x] All existing tests pass

#### Checklist

- [x] Self-review completed
- [x] Code is well-commented

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -22,6 +22,7 @@ const (
 	defaultPostgresBulkBatchSize    = 20000
 	defaultPostgresBulkBatchTimeout = time.Duration(30 * time.Second)
 	defaultPostgresBulkBatchBytes   = int64(80 * 1024 * 1024) // 80MiB
+	defaultPostgresBulkCopyWorkers  = 8
 )
 
 func Load() error {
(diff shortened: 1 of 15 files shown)


cmd/config/config.go
```diff
@@ -174,8 +175,21 @@ func applyPostgresBulkBatchDefaults(batchCfg *batch.Config) {
...
 		batchCfg.MaxBatchBytes = defaultPostgresBulkBatchBytes
+		// Split the batch bytes budget across the send workers: the max queue
+		// bytes semaphore holds both in-flight and accumulating batches, so a
+		// single large default batch (80MiB vs the 100MiB default queue) would
+		// only ever allow one batch in flight and the send workers could never
+		// run concurrently on tables with large rows. Splitting keeps the
+		// memory footprint unchanged while letting up to SendConcurrency
+		// batches circulate.
+		if batchCfg.SendConcurrency > 1 {
+			batchCfg.MaxBatchBytes = defaultPostgresBulkBatchBytes / int64(batchCfg.SendConcurrency)
```
tsg, line 189: Flagged by my agent: this `batchCfg` is also used in `snapshot_and_replication` mode. For replication mode it's not safe because they need to be in order.

So this has to be applied only for the snapshot mode.

tsg: Second comment, might make sense to also split `max_queue_bytes` with the same logic.

tsg (approved):

PR #996, opened 2026-07-11
Add target CREATEDB preflight check for snapshot DB creation

lghuy05:
#### Description

Adds a new access preflight check for `pgstream check` to verify that the target Postgres role has the `CREATEDB` privilege when snapshot schema restore is configured with `created_target_db: true`.

##### Related Issue(s)

- Relate to #897 

#### Type of Change

- [ ] 📚 Documentation update
- [ ] 🔨 Build/CI changes
- [ ] 🧹 Code cleanup

#### Changes Made

- Added `TargetCreateDBPrivilegeCheck` to the access preflight checks
- Wired the check to run only when snapshot dump/restore is configured with `create_target_db: true`
- Added unit tests for the new check and builder coverage for the conditional access check registration

#### Testing

- [x] All existing tests pass

#### Checklist

- [x] Self-review completed
- [ ] Code is well-commented

#### Additional Notes

This change only applies to the snapshot schema restore path for Postgres targets. The new check validates the current target session role by querying `pg_roles` for `rolcreatedb`, and returns an actionable finding when the privilege is missing.


lghuy05:
I will resolve the conflict in a few days

kvch:
How do you handle if the database pgstream is supposed to create is missing? It seems to me that you are using the connection string that contains the database to be created.

lghuy05:
Merged lastest `main` into this branch and resolved the conflicts
@kvch  thanks for the catch,I traced the restore path and saw that `RunPGRestore` removes the database name from the target connection string when `Create` is true, so `pg_restore` can connect before creating the target database.
I will update the preflight check to use the same behavior

pkg/stream/preflight/builder.go
```diff
@@ -132,7 +133,51 @@ func BuildAccessChecks(cfg *stream.Config) ([]Check, CleanupFunc) {
...
+func removeDatabaseFromConnectionString(url string) (string, error) {
+	pgCfg, err := postgres.ParseConfig(url)
+	if err != nil {
+		return "", err
+	}
+	if pgCfg.Database == "" || pgCfg.Database == "postgres" {
+		return url, nil
+	}
+
+	return strings.ReplaceAll(url, "/"+pgCfg.Database, "/"), nil
```
kvch: We already have a `removeDatabaseFromConnectionString` in `internal/postgres` package. Could you please export that and reuse it here?

pkg/stream/preflight/access.go
```diff
@@ -94,6 +102,14 @@ WHERE c.relkind = 'S'
 ORDER BY table_schema, table_name, sequence_schema, sequence_name
 `
 
+const targetCreateDBPrivilegeQuery = `
+SELECT 
+	current_user,
+	rolcreatedb
```
kvch: Superusers can also create databases, but afaik they do not have create table privileges in this table. Please make sure that superusers do not produce false negatives for this check.

pkg/stream/preflight/builder.go
```diff
@@ -132,7 +133,51 @@ func BuildAccessChecks(cfg *stream.Config) ([]Check, CleanupFunc) {
...
+				})
+				return checks, joinCleanups(cleanups)
+			}
+			target := postgres.NewLazyConn(targetURL)
+			checks = append(checks, &TargetCreateDBPrivilegeCheck{Target: target.Acquire})
+			cleanups = append(cleanups, target.Close)
+		}
+	}
+
+	return checks, func(ctx context.Context) error {
```
kvch: The function is the reimplementation of `joinCleanups`. Please use the existing helper.

pkg/stream/preflight/builder.go
lghuy05: yes

pkg/stream/preflight/builder.go
lghuy05: Thank you!

pkg/stream/preflight/access.go
lghuy05: Thanks!

kvch (approved):

PR #999, opened 2026-07-13
Add support for "schema only" tables

tsg:
#### Description

This adds the ability to copy only the schema of some tables. This applies to both the snapshot and the incremental/stream mode. They behave like "included" tables, but never copy any data.

Example (`schema_only_tables` are the new options):

```
source:
  postgres:
    url: "postgres://user:pass@source:5432/mydb"
    mode: snapshot_and_replication
    snapshot:
      mode: full
      tables: ["*"]
      # only the schema is snapshotted for these, no data is copied
      schema_only_tables: ["audit_log", "reports.*"]

target:
  postgres:
    url: "postgres://user:pass@target:5432/mydb"

modifiers:
  filter:
    include_tables: ["*"]
    # DDL (schema changes) for these tables keep replicating,
    # but their data (insert/update/delete) events are skipped
    schema_only_tables: ["audit_log", "reports.*"]
```

#### Type of Change

- [ ] 📚 Documentation update
- [ ] 🔨 Build/CI changes
- [ ] 🧹 Code cleanup

#### Changes Made

-
-
-

#### Testing

- [ ] All existing tests pass

#### Checklist

- [x] Self-review completed
- [ ] Code is well-commented


kvch:
I see in your commits that you save if the table schema was already replicated to the target:
> record schema-only tables as snapshot requests so restarts skip the redundant schema dump+restore instead of re-running it every time

It seems to me if the user decides to replicate data from the table and moves the table from `schema_only_tables` to `tables`, the data is not going to be replicated because `pgstream` considers the table replication complete. We need to distinguish the two replication modes in order to capture what was replicated to the destination exactly.

Another question: are we deliberately copying sequences to the target? It seems contradictory with the goal of the feature.

Furthermore, `schema_only_tables` are included in the `check` command's `SELECT` privilege checking. Getting the schema of a table does not require `SELECT` privilege on the table. So it might produce false errors if the role does not have the privilege.

tsg:
> It seems to me if the user decides to replicate data from the table and moves the table from schema_only_tables to tables, the data is not going to be replicated because pgstream considers the table replication complete. We need to distinguish the two replication modes in order to capture what was replicated to the destination exactly.

I went for adding a `mode` column to the pgstream snapshots schema (`data` | `schema-only`). This way we can know what mode was configured at the previous run.

>Another question: are we deliberately copying sequences to the target? It seems contradictory with the goal of the feature.

I guess it's somewhat debatable if sequences should be part of the schema or not. I'm fine either way.

> Furthermore, schema_only_tables are included in the check command's SELECT privilege checking. Getting the schema of a table does not require SELECT privilege on the table. So it might produce false errors if the role does not have the privilege.

Seems that we do need `SELECT` privileges for the pg_dump part, even with `--schema-only`:

> pg_dump takes an ACCESS SHARE lock on every table in its dump set even with --schema-only, and that lock requires SELECT privilege. I verified it empirically on postgres:17 with a role that had schema USAGE but no table SELECT:

internal/postgres/schema_table_map.go
```diff
@@ -53,12 +60,35 @@ func (t SchemaTableMap) GetSchemaTables(schema string) map[string]struct{} {
...
+	merged := make(map[string]struct{}, len(tables)+len(t[wildcard]))
+	maps.Copy(merged, tables)
+	maps.Copy(merged, t[wildcard])
+	return merged
+}
+
+// ValidateWildcardSchema returns an error when the wildcard schema entry lists
+// anything other than the wildcard table ("*.*"): the snapshot generators
+// can't resolve a specific table name across all schemas.
+func (t SchemaTableMap) ValidateWildcardSchema() error {
```
kvch, line 79: This function should be the only way we get this info. But at the moment we have 2 inline duplications of the same functionality:

  1. `pkg/snapshot/generator/postgres/schema/pgdumprestore/snapshot_pg_dump_restore_generator.go`, in `CreateSnapshot`:
```golang
  if tables, found := schemaOnly[wildcard]; found && (len(tables) != 1 || tables[0] != wildcard) {
      return fmt.Errorf("wildcard schema must be used with wildcard table, got %q", tables)
  }
```
  2. `pkg/snapshot/generator/postgres/tablefinder/pg_snapshot_table_finder.go`, in `CreateSnapshot`:
```golang
  if tablenames, found := ss.SchemaOnlyTables[wildcard]; found && (len(tablenames) != 1 || tablenames[0] != wildcard) {
      return fmt.Errorf("wildcard schema must be used with wildcard table, got %q", tablenames)
  }
```

Note this file also has the similar pattern for `ss.SchemaTables[wildcard]` a few lines down.

You could replace the inline checks with a call to the shared validator so the message and logic can't drift. Convert the `SchemaOnlyTables` `map[string][]string` to a `pglib.SchemaTableMap` (or add a small `[]string/map[string][]string`-based helper next to `ValidateWildcardSchema` if that's cleaner given the call sites) and call the shared method. Plus adjust the tests, given that the error messages slightly differ.

kvch (approved):
I had a minor comment about code duplication for validating wildcard schemas. This part of the code got more complicated, so I would prefer to keep the logic in a single function, and call it everywhere else.

PR #882, opened 2026-06-10
Add new healthcheck endpoints: `/health` and `/ready`

kvch:
#### Description

This PR adds an opt-in HTTP server exposing readiness and liveness probes to the commands `run` and `snapshot`:
* `/health` (liveness, always 200)
* `/ready` (readiness) for monitoring a running pgstream process

Both endpoints respond with JSON `{"status":"ok","version":"..."}`.

`/ready` runs a 2s-timeout SELECT 1 ping query against the source postgres database
when configured, returning 503 with failure detail on error.

For kafka sources or when no source is wired up it behaves like `/health`.

#### Type of Change

Please select the relevant option(s):

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📚 Documentation update
- [ ] 🔧 Refactoring (no functional changes)
- [ ] ⚡ Performance improvement
- [ ] 🧪 Test coverage improvement
- [ ] 🔨 Build/CI changes
- [ ] 🧹 Code cleanup

#### Changes Made

- Added opt-in HTTP health server with `/health` and `/ready` endpoints
- `/ready` pings the source Postgres database with a 2s timeout when configured
- Merged upstream `--upgrade`/`--init` options support into `cmd/run_cmd.go`

#### Testing

- [ ] Unit tests added/updated
- [ ] Integration tests added/updated
- [ ] Manual testing performed
- [ ] All existing tests pass

#### Checklist

- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Code is well-commented
- [x] Documentation updated where necessary

PR #1000, opened 2026-07-14
Add pipeline phase visibility via `/status` and OTel gauge

### LINKED ISSUE #984: Add phase visibility for `snapshot_and_replication` mode
### Problem

In `snapshot_and_replication` mode, there's no queryable way to tell whether the process is doing the initial snapshot or logical replication. The only signal is the ordered log sequence: initial snapshot enabled → snapshot generation completed → logical replication started.

`pgstream status` doesn't help: it's a static config/init check against Postgres (not the running process), and its Status struct has no phase field.

### Proposal

Expose the current phase (snapshot | replication) as a queryable signal. Preferred: a /status endpoint on the existing health server (cmd/health_cmd.go:32), and OTel phase gauge, or enriching pgstream status to read slot active/confirmed_flush_lsn.

### Requirements

- [ ] Operator can tell snapshot vs streaming phase without reading logs.
- [ ] Signal flips at logical replication started.
- [ ] Works for all pgstream modes, not just `snapshot_and_replication`

Shaurya2k06:
#### Description

Adds queryable pipeline phase visibility so operators can tell whether a running pgstream process is in the initial **snapshot** or **replication** (streaming) phase without scraping logs.

In `snapshot_and_replication` mode, the only signal today is log order: `initial snapshot enabled` → `snapshot generation completed` → `logical replication started`. `pgstream status` is a static config/init check against Postgres and has no phase field.

This PR exposes phase via:
1. **`GET /status`** on the existing health server (alongside `/health` and `/ready`)
2. **OTel observable gauge** `pgstream.pipeline.phase`

The phase flips to `replication` when logical replication starts (after successful `StartReplication` / `StartReplicationFromLSN`), matching the existing log signal.

##### Related Issue(s)

- Closes #984

#### Type of Change

- [x] 📚 Documentation update
- [ ] 🔨 Build/CI changes
- [ ] 🧹 Code cleanup

#### Changes Made

- Add `internal/phase.Tracker` — thread-safe atomic tracker with `snapshot` and `replication` phases
- Extend health server with `GET /status` and `WithPhaseProvider`; response includes `phase` field
- Wire tracker from `run` / `snapshot` commands into health server and stream pipeline
- Set phase at lifecycle boundaries in postgres, snapshot-only, and kafka listeners:
  - `snapshot_and_replication`: `snapshot` at start of `snapshotAndListen`, `replication` after `StartReplicationFromLSN` succeeds
  - `replication` only: `replication` after `StartReplication` succeeds
  - `pgstream snapshot`: `snapshot` when snapshot `Listen` starts
  - Kafka source: `replication` when kafka `Listen` starts
- Register OTel `Int64ObservableGauge` `pgstream.pipeline.phase` (value `1` with `phase` attribute)
- Document `/status` in `docs/configuration.md` and `config_template.yaml`
- Document `pgstream.pipeline.phase` in `docs/Observability.md`

**Out of scope (per issue discussion):** enriching CLI `pgstream status` — it runs as a separate process and cannot observe in-memory pipeline state; slot/LSN heuristics would be unreliable.

#### Before / After

| Signal | Before | After |
|--------|--------|-------|
| `GET /status` | 404 (endpoint did not exist) | `{"status":"ok","version":"...","phase":"snapshot"\|"replication"\|""}` |
| `pgstream status` | Static init/config check, no phase | Unchanged |
| OTel metrics | No pipeline phase metric | `pgstream.pipeline.phase{phase="snapshot"\|"replication"} = 1` |
| Logs | Only signal for phase transitions | Unchanged (still emitted) |

#### Testing

  - `internal/phase/phase_test.go` — tracker get/set, nil-safe
  - `internal/health/health_test.go` — `/status` handler with/without provider, phase transitions
  - `pkg/wal/listener/postgres/wal_pg_listener_test.go` — phase flips at `StartReplication` / `StartReplicationFromLSN`
  - `pkg/stream/phase_metrics_test.go` — `registerPhaseMetric` via OTel `ManualReader`
  - `pkg/stream/integration/phase_status_integration_test.go` — polls `GET /status` through snapshot → replication on a real pg2pg `snapshot_and_replication` run; verifies post-snapshot insert replicates to target
  - Integration test exercises live HTTP polling of `/status` during snapshot and replication (automated equivalent of manual `curl`)
- [x] All existing tests pass
  - `make test` passes
  - Integration test passes locally with OrbStack

#### Checklist

- [x] Self-review completed
- [x] Code is well-commented

#### Additional Notes

- Health `/status` is opt-in via existing `instrumentation.health.enabled` (`PGSTREAM_HEALTH_CHECK_ENABLED`); no new config knob added
- OTel gauge uses the existing metrics endpoint configuration
- Phase is empty (`""`) until the first lifecycle transition
- Existing `/health` and `/ready` behavior is unchanged
- Mirrors patterns from health endpoints (#882) and observable gauges (`pgstream.replication.lag`, dropped-queries counter #959)


cmd/run_cmd.go
```diff
@@ -71,13 +72,14 @@ func run(ctx context.Context) error {
...
-	stopHealth, err := startHealthServer(ctx, stdLogger, streamConfig.SourcePostgresURL())
+	phaseTracker := phase.NewTracker()
+	stopHealth, err := startHealthServer(ctx, stdLogger, streamConfig.SourcePostgresURL(), phaseTracker)
 	if err != nil {
 		return err
 	}
 	defer stopHealth()
 
-	return stream.Run(ctx, stdLogger, streamConfig, initFlag, provider.NewInstrumentation("run"), opts...)
+	return stream.Run(ctx, stdLogger, streamConfig, initFlag, provider.NewInstrumentation("run"), phaseTracker, opts...)
```
kvch: Could you please turn the phaseTracker into an `InitOption` for `stream.Run`?

cmd/snapshot_cmd.go
```diff
@@ -44,13 +45,14 @@ func snapshot(ctx context.Context) error {
...
-	stopHealth, err := startHealthServer(ctx, stdLogger, streamConfig.SourcePostgresURL())
+	phaseTracker := phase.NewTracker()
+	stopHealth, err := startHealthServer(ctx, stdLogger, streamConfig.SourcePostgresURL(), phaseTracker)
 	if err != nil {
 		return err
 	}
 	defer stopHealth()
 
-	return stream.Snapshot(ctx, stdLogger, streamConfig, provider.NewInstrumentation("snapshot"))
+	return stream.Snapshot(ctx, stdLogger, streamConfig, provider.NewInstrumentation("snapshot"), phaseTracker)
```
kvch: Could you please turn the `phaseTracker` into an optional parameter for `stream.Snapshot`?

kvch:
You can ignore the coverage failure, but please fix the issues reported by the linter.

pkg/stream/phase_metrics.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return fmt.Errorf("creating pipeline phase gauge: %w", err)
+	}
+
+	_, err = instrumentation.Meter.RegisterCallback(func(_ context.Context, o metric.Observer) error {
+		p := tracker.Get()
+		if p == "" {
+			return nil
+		}
+		o.ObserveInt64(gauge, 1, metric.WithAttributes(attribute.String("phase", string(p))))
```
kvch: I've tested this change, and the metric seems a bit counter-intuitive in this format. The gauge `pgstream.pipeline.phase` only observes the currently-active phase. So the phase flips from `snapshot` to `replication`, the `{phase="snapshot"}` series stops being reported and goes stale rather than dropping to `0`, which makes `sum by (phase)` and alerting queries ambiguous.

Let's switch to the standard enum-as-gauge pattern (like kube-state-metrics' `kube_pod_status_phase`): emit every known phase on each collection, `1` for the active one and `0` for the rest:

| Tracker state | `{phase="snapshot"}` | `{phase="replication"}` |
|---|---|---|
| empty (no phase set yet) | *(not emitted)* | *(not emitted)* |
| snapshot | **1** | 0 |
| replication | 0 | **1** |

**File:** `pkg/stream/phase_metrics.go`, in `registerPhaseMetric`'s `RegisterCallback` callback.

**Current:**

  ```go
  _, err = instrumentation.Meter.RegisterCallback(func(_ context.Context, o metric.Observer) error {
      p := tracker.Get()
      if p == "" {
          return nil
      }
      o.ObserveInt64(gauge, 1, metric.WithAttributes(attribute.String("phase", string(p))))
      return nil
  }, gauge)
  ```

  **Change to:**

  ```go
  _, err = instrumentation.Meter.RegisterCallback(func(_ context.Context, o metric.Observer) error {
      current := tracker.Get()
      if current == "" {
          return nil // no phase entered yet — emit nothing
      }
      for _, p := range []phase.Phase{phase.Snapshot, phase.Replication} {
          var v int64
          if p == current {
              v = 1
          }
          o.ObserveInt64(gauge, v, metric.WithAttributes(attribute.String("phase", string(p))))
      }
      return nil
  }, gauge)
  ```

Please also update the gauge's `metric.WithDescription(...)` string: change "reports 1 with attribute phase=snapshot|replication" to `"Reports 1 for the active pipeline phase and 0 for the others (phase=snapshot|replication)"`.

pkg/wal/listener/postgres/wal_pg_listener.go
```diff
@@ -77,18 +79,30 @@ func WithInitialSnapshot(sg snapshotGenerator) Option {
...
 // Listen starts the subscription process to listen for updates from PG.
 func (l *Listener) Listen(ctx context.Context) error {
 	if l.snapshotGenerator != nil {
 		if err := l.snapshotAndListen(ctx); err != nil {
 			l.logger.Error(err, "pg snapshot and listen")
 			return err
 		}
+		// snapshotAndListen already entered the listen loop; only reached on
+		// clean return (context cancel) or if snapshot path returned early.
+		return nil
```
kvch, line 99: Could you please explain the reasoning behind this change?

pkg/stream/stream_snapshot.go
```diff
@@ -7,14 +7,17 @@ import (
...
 	"github.com/xataio/pgstream/pkg/otel"
 	snapshotlistener "github.com/xataio/pgstream/pkg/wal/listener/snapshot"
 	snapshotbuilder "github.com/xataio/pgstream/pkg/wal/listener/snapshot/builder"
 	"golang.org/x/sync/errgroup"
 )
 
-func Snapshot(ctx context.Context, logger loglib.Logger, config *Config, instrumentation *otel.Instrumentation) error {
+// Snapshot performs a one-time data snapshot. This call is blocking.
+// phaseTracker is optional; when provided it is set to snapshot for the duration.
+func Snapshot(ctx context.Context, logger loglib.Logger, config *Config, instrumentation *otel.Instrumentation, phaseTracker ...*phase.Tracker) error {
```
kvch: I am sorry, I realize now that my wording was a bit misleading. I did not mean to pass a variadic argument. Rather as an `InitOption` like in the replication case (`Run`).

New signature:
 ```go
func Snapshot(ctx context.Context, logger loglib.Logger, config *Config, instrumentation *otel.Instrumentation, opts ...InitOption) error {
  ```
Replace the variadic-unpacking block:
```go
var tracker *phase.Tracker
if len(phaseTracker) > 0 {
    tracker = phaseTracker[0]
}
```

with the same accessor `Run` uses:

```go
tracker := config.GetInitConfig(opts...).PhaseTracker
```

kvch (changes requested):
Thank your for getting back to the PR. I've posted a few notes.

pkg/stream/stream_snapshot.go
Shaurya2k06: Got it! Thanks for the feedback.

pkg/stream/phase_metrics.go
Shaurya2k06: We can do that. I'll update accordingly.

pkg/wal/listener/postgres/wal_pg_listener.go
Shaurya2k06: The old fallthrough was unreachable in the current code path, because snapshotAndListen() always ends by calling listen() , and listen() only returns a non-nil error or ctx.Err(), not nil. So the explicit return nil isn't really a behavior fix, it's mainly about safety, it makes it obvious that once the snapshot path is used, the function should stop there and not continue into the normal replication startup path.It also prevents a future refactor from accidentally making the old fallthrough reachable and then calling StartReplication() twice on the same listener flow.

How do you think this should be dealt with? Again, It's guard-railing/documentation, not a bug fix.

kvch (approved):
Thank you!

PR #1008, opened 2026-07-17
feat: add prometheus endpoint

### LINKED ISSUE #903: Add Prometheus `/metrics` endpoint
**Is your feature request related to a problem? Please describe.**

`pgstream` currently exports metrics only via OTLP gRPC push (`pkg/otel/otel_provider.go`), which forces every deployment to run an OpenTelemetry collector just to read basic metrics like replication lag. Pull-based scraping is the default in Kubernetes and the common case for operators running `pgstream` as a service.

**Describe the solution you'd like**

Expose a Prometheus `/metrics` endpoint on the existing health HTTP server (`internal/health/health.go`) using the OTel Prometheus exporter, so the same instruments feed both OTLP push and pull-based scraping. Gate it behind config (e.g. `instrumentation.metrics.prometheus.enabled`/`endpoint` path), defaulting off to preserve current behavior.

pauliyobo:
#### Description
This PR allows pgstream to expose a configurable Prometheus endpoint, served from the existing health server, so metrics can be scraped directly via the OTel Prometheus exporter without requiring an OTLP collector.

#### Related Issue(s)
closes #903

#### Type of Change

- [ ] 📚 Documentation update
- [ ] 🔨 Build/CI changes
- [ ] 🧹 Code cleanup

#### Changes Made

- Add `instrumentation.metrics.prometheus` config (YAML and env vars) to enable/configure a Prometheus scrape endpoint.
- Serve the endpoint from the health server (default path `/metrics`, returns 404 when disabled).
- Wire the OTel Prometheus exporter into the meter provider independently of the OTLP exporter, so Prometheus works standalone without an OTLP endpoint configured.
- Update configuration and observability docs for the new option.

#### Testing

- [x] All existing tests pass

#### Checklist

- [x] Self-review completed
- [ ] Code is well-commented

#### Additional Notes
I have made the endpoint configurable as well. Is that fine, or should I leave /metrics?


internal/health/health.go
```diff
@@ -114,6 +119,14 @@ func (s *Server) Listen() error {
 	mux.HandleFunc("/ready", s.handleReady)
 	mux.HandleFunc("/status", s.handleStatus)
 
+	if s.isPrometheusEnabled() {
+		endpoint := DefaultPrometheusEndpoint
+		if promEndpoint := s.metricsConfig.Prometheus.Endpoint; promEndpoint != "" {
```
kvch, line 124: Please validate the endpoint format.

cmd/health_cmd.go
```diff
@@ -60,7 +66,11 @@ func startHealthServer(ctx context.Context, logger loglib.Logger, sourcePostgres
 		opts = append(opts, health.WithReadinessCheck(pool.Ping))
 	}
 
-	srv := health.NewServer(*cfg, opts...)
+	var metricsCfg *otel.MetricsConfig
+	if instrumentationCfg.Metrics != nil {
```
kvch, line 70: It seems to me that if the health server is not enabled, but prometheus is enabled, the endpoint is not going to work.

pauliyobo: Yeah, that's correct. This is because of the dependency the metrics endpoint has on the health server. this was also documented. What would be preferred in this case?

kvch: Could you please add a warning when this happens? It is possible that the user misconfigures it.

kvch (approved):

PR #1038, opened 2026-07-27
Fix replicated DELETEs on user-defined enum PKs

### LINKED ISSUE #1037: Replicated DELETEs on tables with a user-defined enum PK column fail to encode (bulk-delete coalescing binds Go slices for enum array OIDs)
### Summary

In `snapshot_and_replication` mode, replicated `DELETE` events on a table whose primary key (replica identity) contains a **user-defined enum column** always fail to encode, and the failing query poisons its whole write batch. The v1.2.3 enum fix (#1035) covered the snapshot bulk-ingest COPY path, but the batch writer's coalesced-DELETE path has the same underlying problem.

### Environment

- pgstream v1.2.3, `source.postgres.mode: snapshot_and_replication`, wal2json
- Target: `bulk_ingest.enabled: true`, `batch.ignore_send_errors: true`

### Error

```
ERR logger.go:51 > executing sql query error.message="failed to encode args[1]: unable to encode []interface {}{\"BiteScan\"} into text format for unknown type (OID ...)"
```

### Reproduction

```sql
CREATE TYPE asset_kind AS ENUM ('BiteScan', 'UpperJawScan', 'LowerJawScan');
CREATE TABLE asset (
    export_id uuid NOT NULL,
    kind asset_kind NOT NULL,
    number int NOT NULL,
    PRIMARY KEY (export_id, kind, number)
);
```

Run pgstream with a postgres target, then on the source:

```sql
INSERT INTO asset VALUES (gen_random_uuid(), 'BiteScan', 1);
DELETE FROM asset WHERE kind = 'BiteScan';
```

The INSERT replicates; the DELETE fails to encode. A single-PK table with an enum PK (`PRIMARY KEY (kind)`) fails the same way via the `ANY($1::asset_kind[])` variant.

### Root cause

`BatchWriter.buildCoalescedQueries` routes every run of consecutive same-table DELETEs — even a run of length 1 — through `dmlAdapter.buildBulkDeleteQuery` (`pkg/wal/processor/postgres/postgres_batch_writer.go:204`).

For composite PKs, `buildBulkDeleteCompositePK` (`pkg/wal/processor/postgres/postgres_wal_dml_adapter_bulk.go`) emits:

```sql
DELETE FROM t WHERE (export_id, kind, number) IN (SELECT * FROM unnest($1::uuid[], $2::asset_kind[], $3::int4[]))
```

binding one Go `[]any` per PK column. pgx's `pgtype.Map` has no codec registered for the database-specific enum **array** OID, so encoding the `kind` slice fails client-side (`unable to encode []interface {}{...} into text format for unknown type`). Scalar enum values are plain strings and encode fine into unknown OIDs, which is why INSERT/UPDATE work — only the bulk-DELETE path binds slices.

`buildBulkDeleteSinglePK` has the same issue via `pgArrayType` (`WHERE kind = ANY($1::asset_kind[])`).

### Impact amplification

The encode error is a client-side error, not one of the recognized `pglib` error types, so `BatchWriter.isInternalError` returns `true` and the drop-one-query-and-retry-the-rest logic in `execQueries` is skipped — the whole send fails. With `ignore_send_errors: true`, the **entire batch is dropped**, including co-batched writes to unrelated tables, so the target silently diverges beyond the deleted rows themselves. Since the coalescing is unconditional, every DELETE on such a table hits this; there is no configuration workaround.

### Suggested fix

Either of:

1. Bind identity values for enum columns as `text[]` and cast, e.g. `$2::text[]::asset_kind[]` (or a per-column cast inside the unnest SELECT: `SELECT c1, c2::asset_kind, c3 FROM unnest($1::uuid[], $2::text[], $3::int4[]) AS u(c1,c2,c3)`). wal2json already delivers enum identity values as strings. `schemaInfo.enumColumns` (used by the COPY fix in #1035) already identifies which columns need this.
2. Register user-defined enum types (and their array OIDs) in the pgx type map, as the schema information to do so is already being queried.

Happy to provide more logs or test the fix — we hit this in production replication with pgstream v1.2.3.

kvch:
#### Description

Replicated `DELETE`s on tables whose replica identity contains a user-defined enum column always failed to encode, and the failure took down the whole batch. The enum fix (#1035) covered the snapshot bulk-ingest COPY path only.

Identity values for enum columns are now bound as `text[]` and cast back on the target. Cast targets are resolved from the **target catalog** (the enum itself, through the array element type and the domain base type), not from the column type carried by the replication stream. Casting to the column's declared type is not equivalent: the enum comparison operators are polymorphic over `anyenum`, which does not accept a domain, so `dom = dom` has no operator while `dom::the_enum = the_enum` does.

The error was also misclassified. Being client-side it carries no SQLSTATE, so `isInternalError` treated it as internal, skipping the drop-one-query-and-retry path — with `ignore_send_errors: true` the entire batch was dropped, including co-batched writes to unrelated tables. It is now a typed `ErrValueEncoding`, non-internal in the writer and non-retriable in the retrier.

##### Related Issue(s)

  - Fixes #1037

  #### Type of Change

  - [x] 📚 Documentation update
  - [ ] 🔨 Build/CI changes
  - [ ] 🧹 Code cleanup

  #### Changes Made

  - `enumTableColumnsQuery` now returns, per enum-resolving column, the quoted schema-qualified enum type plus `isArray`/`isDomain`; `schemaInfo.enumColumns` carries that instead of a bare name set.
  - Bulk DELETE emits the shape each column needs, all verified against a live PostgreSQL: `= ANY($1::text[]::enum[])` for a scalar enum (keeps the index — `Index Cond: (col = ANY (...))`), `IN (SELECT unnest($1::text[])::enum[])` for an enum array, and an additional
  column-side cast for a domain over an enum. Tables with no enum identity column emit byte-identical SQL to before.
  - Fixes a pre-existing gap: deletes on a domain-over-enum column failed on **every** path, including the per-row form, with `operator does not exist: dom = dom`.
  - New `pglib.ErrValueEncoding`, recognised in `MapError` after the `PgError` switch (a client-side failure can never carry a SQLSTATE, and checking first could shadow a real one, since postgres embeds row values in messages). `Details` keeps only pgx's prefix — pgx
  formats the offending argument with `%#v`, which for a bulk delete is every identity value in the batch, and that string reaches logs and OTel spans.
  - Added to `isRetriableError`. The default policy sets neither `MaxRetries` nor `MaxElapsedTime`, so a deterministic encode failure would otherwise retry forever with a connection-pool reset per attempt, and the drop-one-query recovery would never be reached.

  #### Testing

  - [x] All existing tests pass

  #### Checklist

  - [x] Self-review completed
  - [x] Code is well-commented

### DIFF
--- a/internal/postgres/errors.go
+++ b/internal/postgres/errors.go
@@ -112,6 +112,18 @@ func (e *ErrFeatureNotSupported) Error() string {
 	return fmt.Sprintf("feature not supported: %s", e.Details)
 }
 
+type ErrValueEncoding struct {
+	Details string
+}
+
+func (e *ErrValueEncoding) Error() string {
+	return fmt.Sprintf("value encoding: %s", e.Details)
+}
+
+const encodeArgsErrFragment = "failed to encode args["
+
+var encodeArgsErrDetails = regexp.MustCompile(`^(.*?failed to encode args\[\d+\]).*?(for [^:]+)`)
+
 // MapError maps a Postgres error to a more specific error type
 func MapError(err error) error {
 	if pgconn.Timeout(err) {
@@ -206,5 +218,21 @@ func MapError(err error) error {
 		}
 	}
 
+	// A client-side encoding failure is recognized last, and only when nothing
+	// carried a SQLSTATE: it never reaches the server, so it can never be a
+	// PgError. Checking it earlier would let a genuine server error whose
+	// message happens to quote the fragment — postgres embeds row values in
+	// messages, and pgstream is often pointed at tables holding log text — be
+	// misclassified, changing how the writer and the retrier handle it.
+	if err != nil && pgErr == nil && strings.Contains(err.Error(), encodeArgsErrFragment) {
+		return &ErrValueEncoding{Details: encodingErrDetails(err.Error())}
+	}
+
 	return err
+}
+func encodingErrDetails(msg string) string {
+	if m := encodeArgsErrDetails.FindStringSubmatch(msg); m != nil {
+		return m[1] + " " + m[2]
+	}
+	return encodeArgsErrFragment
 }
--- a/internal/postgres/errors.go
+++ b/internal/postgres/errors.go
@@ -5,6 +5,7 @@ package postgres
 import (
 	"errors"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/jackc/pgx/v5"
@@ -230,6 +231,7 @@ func MapError(err error) error {
 
 	return err
 }
+
 func encodingErrDetails(msg string) string {
 	if m := encodeArgsErrDetails.FindStringSubmatch(msg); m != nil {
 		return m[1] + " " + m[2]
--- a/internal/postgres/errors_test.go
+++ b/internal/postgres/errors_test.go
@@ -4,11 +4,30 @@ package postgres
 
 import (
 	"errors"
+	"fmt"
 	"testing"
 
+	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgconn"
+	"github.com/jackc/pgx/v5/pgtype"
 	"github.com/stretchr/testify/require"
 )
+
+// TestMapError_realPgxEncodeError guards the message matching in MapError
+// against pgx changing its wrapper text: the error here is produced by pgx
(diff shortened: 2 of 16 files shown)


kvch:
Forgot to push the changes before the bot triggered.

PR #1049, opened 2026-07-29
Added tests and docs for snapshot+stream from a read replica

tsg:
#### Description

This adds an integration test for doing snapshot+streaming from a read-replica instead of from the primary. Also adds a docs page to have the procedure documented.

The only code change besides the tests is a new tiny feature: `pgstream init` gets a `--slot-only` flag. This is the equivalent of running:

```
SELECT pg_create_logical_replication_slot('pgstream_<db>_slot', 'wal2json');
```

And it simplifies a bit the setup instructions. 

#### Type of Change

- [x] 📚 Documentation update
- [ ] 🔨 Build/CI changes
- [ ] 🧹 Code cleanup

### DIFF
--- a/cmd/root_cmd.go
+++ b/cmd/root_cmd.go
@@ -70,6 +70,12 @@ func Prepare() *cobra.Command {
 	initCmd.Flags().Bool("migrations-only", false, "Whether to only run the initialization database migrations")
 	initCmd.Flags().Bool("slot-only", false, "Whether to only create the replication slot, without running the database migrations. Useful when creating the slot on a read replica, where the migrations cannot be applied")
 	initCmd.Flags().Bool("upgrade", false, "Clean up v0.9.x state before initializing (idempotent, safe for repeated use)")
+	// slot-only selects the half of the work the other two flags act on, so
+	// combining them is always a mistake: --migrations-only asks for the
+	// complementary half, and --upgrade only cleans up schema state that
+	// slot-only never touches, which would otherwise be silently ignored
+	initCmd.MarkFlagsMutuallyExclusive("migrations-only", "slot-only")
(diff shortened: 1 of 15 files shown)


geo-ocean (approved):

pkg/stream/integration/pg_replica_integration_test.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+	runStream(t, ctx, cfg)
+
+	targetConn, err := pglib.NewConn(ctx, targetPGURL)
+	require.NoError(t, err)
+	defer targetConn.Close(ctx)
+
+	// the snapshot reads from the standby, including the parallel path that
+	// exports a transaction snapshot and imports it on sibling connections
+	require.Eventually(t, func() bool {
```
kvch, line 80: `fetchReplicaTestRows` and `targetHasColumn` call `require.NoError`, and both are reached from inside `Eventually` closures. A failing `rows.Scan`/`rows.Err` calls `runtime.Goexit` on a non-test goroutine → the result is never sent on `testify`'s channel → the test stalls the full 60s and reports a generic "timeout waiting for snapshot" that masks the real error.

cmd/root_cmd.go
```diff
@@ -68,13 +68,15 @@ func Prepare() *cobra.Command {
 	initCmd.Flags().String("replication-slot", "", "Name of the postgres replication slot to be created by pgstream on the source url")
 	initCmd.Flags().Bool("with-injector", false, "Whether to initialise pgstream with the injector database migrations")
 	initCmd.Flags().Bool("migrations-only", false, "Whether to only run the initialization database migrations")
```
kvch, line 70: `migrations-only` and `slot-only` should be mutually exclusive. Also, `upgrade` and `slot-only` as well.

kvch (approved):
WFG

PR #1056, opened 2026-08-03
Pin the data snapshot to the columns the schema snapshot saw

kvch:
#### Description

  The data snapshot read rows with `SELECT * FROM ONLY <table> WHERE ctid BETWEEN …`. `SET TRANSACTION SNAPSHOT` pins row visibility but not the catalog, and every page range plans its own query, so the column list was re-resolved from the live source catalog throughout the snapshot, while the target's schema stays the same.

This pins each table's read to the columns the schema snapshot captured, so a column added afterwards is simply not read. The `ADD COLUMN` and any backfill arrive over the replication stream, which already starts from an LSN taken before the snapshot.

#### Type of Change
  - [ ] 📚 Documentation update
  - [ ] 🔨 Build/CI changes
  - [ ] 🧹 Code cleanup

  #### Changes Made

  - `pglib.DiscoverTableColumns` captures the source columns of the in-scope tables; the schema snapshot generator runs it immediately **before** `pg_dump` and hangs the result on `snapshot.Snapshot.TableColumns`, which flows down the generator chain to the data snapshot.
  - The data snapshot builds an explicit projection (`SELECT "c1","c2",… FROM ONLY t …`) once per table and reuses it for every page range, instead of `SELECT *`.
  - The pinned list is intersected with the columns the table still has, so a concurrent `DROP COLUMN` skips the column rather than aborting the table over a DDL that replication is about to replay.
  - New `ErrSchemaChangedDuringSnapshot` for the narrow race where a column disappears between resolving the list and planning a page range.

  #### Testing

  - [x] All existing tests pass

### DIFF
--- a/internal/postgres/pg_utils.go
+++ b/internal/postgres/pg_utils.go
@@ -75,6 +75,11 @@ func QuoteIdentifier(s string) string {
 	return pq.QuoteIdentifier(s)
 }
 
+// QuoteRawIdentifier always quotes.
+func QuoteRawIdentifier(s string) string {
+	return pq.QuoteIdentifier(s)
+}
+
 // UnquoteIdentifier reverses the quoting applied by QuoteIdentifier. If the
 // string is not a quoted identifier, it is returned as-is. If it is a quoted
 // identifier, the leading and trailing quotes are removed, and any embedded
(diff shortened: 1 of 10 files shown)


kvch:
Split up into multiple prs #1057 and #1058

PR #1083, opened 2026-08-10
Reject masking configurations that mask no characters

kvch:
#### Description

A `custom` mask whose range covers zero characters (`mask_begin` equal to `mask_end`, or `unmask_begin: 0`) leaves values completely unmasked while the config still reads `name: masking`. It looks anonymized in review and ships raw values.

#### Type of Change

- [ ] ✨ New feature
- [x] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Changes Made

- Reject the configuration at construction, pointing at `noop`, which is the express way to pass a column through untouched.
- Classified `masking` as `lossy`.

#### Testing

- [x] All existing tests pass

#### Additional Notes

Breaking for anyone who has a zero-width custom mask configured today; the fix is one line (`noop`, or a real mask range). Part 2 of a stack — targets `pr1-uniqueness-classification`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -230,7 +230,7 @@ jobs:
           output-file: pgstream.image.sbom.spdx.json
 
       - name: Attest container image SBOM
-        uses: actions/attest@v4.2.0
+        uses: actions/attest@v4.2.2
         with:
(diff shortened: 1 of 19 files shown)


kvch:

PR #1084, opened 2026-08-10
Classify the greenmask transformers for uniqueness

kvch:
#### Description

Part 3 of the uniqueness stack. Classification only, no behaviour change.

#### Type of Change

#### Changes Made

- `greenmask_boolean`, `greenmask_choice` and `greenmask_firstname` are `lossy`: they draw from a fixed, small set of values (two booleans, the configured choice list, a name dictionary), so they collide well before a table of any size is exhausted.
- The range and string generators are `not_guaranteed`: how likely a duplicate is depends on the configured bounds rather than on the input.

#### Testing

- [x] All existing tests pass

#### Additional Notes

Targets `pr2-masking-noop-rejection`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -230,7 +230,7 @@ jobs:
           output-file: pgstream.image.sbom.spdx.json
 
       - name: Attest container image SBOM
-        uses: actions/attest@v4.2.0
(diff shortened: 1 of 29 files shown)


kvch:

PR #1087, opened 2026-08-10
Surface uniqueness findings in status and validate rules

kvch:
#### Description

Part 6 of the stack. The non-blocking findings from the previous PR have no way out of the rule parser; this reports them.

#### Type of Change

#### Changes Made

- `TransformationRulesStatus` gained `Warnings`, reported by `pgstream status` and `pgstream validate rules` (including `--json`) alongside the errors.
- Warnings are logged when the pipeline starts.
- `validate rules` distinguishes valid-with-warnings from invalid.

#### Testing

- [x] All existing tests pass

The propagation test was mutation-checked: deleting the assignment makes it fail.

#### Additional Notes

Targets `pr5-unique-index-check`.

### DIFF
--- a/cmd/config/config_yaml.go
+++ b/cmd/config/config_yaml.go
@@ -328,9 +328,10 @@ type TableTransformersConfig struct {
 }
 
 type ColumnTransformersConfig struct {
-	Name              string         `mapstructure:"name" yaml:"name"`
-	Parameters        map[string]any `mapstructure:"parameters" yaml:"parameters"`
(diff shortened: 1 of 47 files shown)


kvch:
Folded into #1086.

The review agent escalated #1086 because \`WithUniquenessEnforcement()\` and \`Warnings()\` had no callers there — the wiring lived here, so #1086 on its own would have merged as dead code and delivered none of the behaviour it described. Splitting the check from the code that invokes it was a bad seam: it leaves a PR that cannot be reviewed on its merits.

#1086 now carries both, at 9 substantive files / 282 lines, still within the gate.

PR #1129, opened 2026-08-31
Copy passthrough during snapshot mode if no transformers are configured

kvch:
#### Description

This PR adds support for snapshotting a table by piping the source's `COPY ... TO STDOUT` straight into the target's `COPY ... FROM STDIN` when the rows do not need to be parsed. When a transformer, filter, injector or sanitizer is configured, we fall back to the previous behaviour.

Previously, we read read rows, decoded every value into Go values via pgx, wrapped them in `wal.Event` structs, passed them through the processor chain to the bulk ingest writer, which re-encoded them into a `COPY`. Both ends were already `COPY`; only the round trip through Go in the middle is removed. I measured improvement on ~17% on a text heavy table. It is a workload-specific improvement, since the saving is largest where decoding allocates most. So it can be higher for some cases.

Screenshot 2026-08-31 at 20-43-23 pgstream benchmarks

Furthermore, #801 (cube) and #1035 (enum) exist because pgx cannot produce those types' binary encodings, which is why `textOnlyCopyTypes` and `needsTextCopy` exist. Server-to-server COPY never touches a pgx codec, so that family of bugs cannot occur on this path.

Builds on the raw COPY stream primitives from #1126.

**Please review commit by commit.**

- **`feat: stream snapshot rows without decoding them`**: the new logic
- **`refactor: make the page range reader a strategy`**: `pg_snapshot_range_snapshotter.go` looks like a new file but is `processTableRange` relocated: three receiver renames and one inlined variable

#### Type of Change

- [ ] 📚 Documentation update
- [ ] 🔨 Build/CI changes
- [ ] 🧹 Code cleanup

#### Changes Made

**Commit 1: the passthrough**

- The two COPYs run either side of an `io.Pipe`, so memory is bounded by the pipe rather than the page range, and each side closes its end with its error so neither can block on the other
- Binary format throughout
- Generated columns are named by neither end, taken from the **target's** catalog since the target is what rejects them. A table whose columns are all generated has nothing COPY can carry and falls back to decoding.

**Commit 2: the strategy extraction**

- Reading a page range becomes an interface with two implementations
- The reader is handed the transaction *runner* rather than a transaction, so one that must wait waits before the source transaction opens rather than holding a source connection idle in transaction
- Construction goes through `newSnapshotGenerator`, which takes the reader as an argument, so no path can leave it unset.

#### Testing

#### Checklist

- [x] Self-review completed
- [x] Code is well-commented
- [x] Documentation updated where necessary (no user-facing config surface)

### DIFF
--- a/pkg/snapshot/generator/postgres/data/pg_snapshot_copy_passthrough.go
+++ b/pkg/snapshot/generator/postgres/data/pg_snapshot_copy_passthrough.go
@@ -1,0 +1,330 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package postgres
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+	"sync/atomic"
+	"time"
+
+	pglib "github.com/xataio/pgstream/internal/postgres"
+	pglibinstrumentation "github.com/xataio/pgstream/internal/postgres/instrumentation"
+	pglibretrier "github.com/xataio/pgstream/internal/postgres/retrier"
+	synclib "github.com/xataio/pgstream/internal/sync"
+	"github.com/xataio/pgstream/pkg/backoff"
+	loglib "github.com/xataio/pgstream/pkg/log"
+	"github.com/xataio/pgstream/pkg/otel"
+	"golang.org/x/sync/errgroup"
+)
+
+var (
+	errMissingCopyPassthroughTarget = errors.New("copy passthrough requires a target postgres url")
+	errUnexpectedCopiedRows         = errors.New("number of rows copied doesn't match the source rows")
+	errChunkAlreadyCommitted        = errors.New("the target committed the page range before the source transaction failed")
+)
+
+// copyPassthroughMover streams a page range from the source's COPY TO
+// STDOUT into the target's COPY FROM STDIN, decoding nothing on the way. It
+// writes to the target itself, so it takes on what the bypassed writer did:
+// the target connection and its retry policy, trigger suppression, and a
+// budget capping concurrent COPYs.
+//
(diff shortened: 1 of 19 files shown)


pkg/snapshot/generator/postgres/data/pg_snapshot_copy_passthrough.go
```diff
@@ -0,0 +1,265 @@
...
+	maxConnections, err := pglib.ConnPoolMaxConnections(cfg.TargetURL, poolOpts...)
+	if err != nil {
+		return nil, fmt.Errorf("resolving copy passthrough target connections: %w", err)
+	}
+
+	var targetConn pglib.Querier
+	if cfg.RetryPolicy.DisableRetries {
+		targetConn, err = pglib.NewConnPool(ctx, cfg.TargetURL, poolOpts...)
+	} else {
+		targetConn, err = pglibretrier.NewQuerier(ctx, cfg.RetryPolicy, func(ctx context.Context) (pglib.Querier, error) {
```
tsg: Do we need retries in this case? From my agent:

> The target pool is wrapped in the retrier, which re-runs the ExecInTx closure on a retriable error. The pipe reader is already partially consumed by then, so the retry sends a truncated COPY and fails with the row count mismatch, then retries again until the backoff gives up. Nothing corrupt lands because the target tx rolls back, but the retry loop burns time and logs misleading warnings. The passthrough should probably disable retries on the target conn and let the existing page range failure abort the table, or retry the whole range from the source.

tsg (approved):

kvch:
The PR has diverged a lot from main. I am putting up new stacked prs, and closing this one.

PR #1148, opened 2026-09-03
Add `lookup_choice` transformer

### LINKED ISSUE #1137: column transformer randomise foreign key
**Is your feature request related to a problem? Please describe.**

We have tables with foreign key columns into "lookup" tables by id that we would like to randomise choosing an id from the lookup table.

**Describe the solution you'd like**

Something like the below, presumably only applicable for columns that are a foreign key:
```
column_transformers:
  lookup_id: 
    name: foreign_key_choice
    parameters:
      generator: random | sequential  # random = random, sequential = just go around the available ids in a loop
      ignore_values: [ ... ]  # optional in case we have some ids we want to  skip
```

**Describe alternatives you've considered**

I currently work around this by using greenmask_integer or greenmask_choice but this requires re-generating config if the foreign key lookup table contents change.

kvch:
#### Description

This PR adds a transformer named `lookup_choice`. It replaces a column value with one read from a column of another table, loaded from the source database once when the pipeline starts. This is the live-table counterpart of `greenmask_choice`, which chooses from a list written into the configuration. The main issue was that `greenmask_choice` and `greenmask_integer` force users to regenerate their YAML whenever the lookup table's contents change.

The name is not `foreign_key_choice`, as the issue proposes, because nothing here queries `pg_constraint` or checks that a foreign key exists.

| Name | Description |
| --- | --- |
| `lookup_table` | The schema-qualified table to read the replacement values from, quoted as written, so `public.countries` means lowercase `countries`. |
| `lookup_column` | The column of that table whose values become the pool to choose from, read once at startup with `IS NOT NULL` and an `ORDER BY`. |
| `generator: deterministic` | Picks the value from a hash of the incoming value, so equal originals stay equal afterwards; `random` picks independently per row and destroys that grouping. |
| `max_values`    | Max number of values read. Default: 100000 |
| `ignore_values: [0, -1]` | Drops those values from the pool after it is read. |
| `postgres_url` | The database to read the lookup table from; omitted here because the PostgreSQL parser fills in the source URL, and only needed when the source is not PostgreSQL. |

Example configuration

```yaml
 transformations:
    table_transformers:
      - schema: public
        table: addresses
        column_transformers:
          country_id:
            name: lookup_choice
            parameters:
              lookup_table: public.countries
              lookup_column: id
              generator: deterministic
              ignore_values: [0, -1]
```

##### Related Issue(s)

- Closes #1137

#### Type of Change

- [ ] 📚 Documentation update
- [ ] 🔨 Build/CI changes
- [ ] 🧹 Code cleanup

#### Testing

- [x] All existing tests pass

#### Checklist

- [x] Self-review completed
- [x] Code is well-commented


thorfi:
@kvch FYI this will need to ensure that lookup table stream is fully updated first (in snapshot mode - replication mode should be OK), if that's not already taken care of, otherwise the lookup will try to add ids that don't yet exist in the target.  Particularly bad if there are parallel table workers going on.

Alternately the lookup could be done in the target, but that's not as nice.

kvch:

kvch:
> FYI this will need to ensure that lookup table stream is fully updated first (in snapshot mode - replication mode should be OK), if that's not already taken care of, otherwise the lookup will try to add ids that don't yet exist in the target. Particularly bad if there are parallel table workers going on.

This is not going to be an issue because pgstream only restores FK constraints after the tables are copied to the target. The lookup table is up to date at that point.

docs/transformers.md
```diff
@@ -1262,6 +1263,73 @@ transformations:
...
+
+⚠️ **The mapping only holds while the lookup column does.** The value is chosen by position in the loaded list, so inserting or deleting a single row in the lookup table — or editing `ignore_values` — remaps almost every input the next time the pipeline starts. Deterministic mode is reproducible across restarts for a **fixed** lookup set; it is not stable across changes to it. New rows in the lookup table are not picked up until a restart.
+
+⚠️ **The whole column is loaded into memory, once per rule.** There is no limit on how many values are read, and no paging, and each column rule runs its own query: three columns reading the same lookup table load it three times. Point this at a lookup table, not at a large one. The read is given 30 seconds, so a locked or unreachable lookup table fails startup instead of hanging it.
```
tsg, line 1303: Should we add a limit for safety? 100K or something like this.

tsg (approved):

PR #1168, opened 2026-09-07
Send the replication batch as one pipelined exchange

siriusfreak:
## The problem

`BatchWriter.execQueries` runs the queries of a batch in one transaction, but it sends them one at a time:

```go
for i, q := range queries {
    if _, err := tx.Exec(ctx, q.sql, q.args...); err != nil {
```

`pgx.Exec` sends a query and then waits for the result. Therefore a batch of N queries costs N network round trips. The batch removes the cost of `BEGIN` and `COMMIT`, but it does not remove the cost of the exchanges.

This is expensive when the target database is far from the client. We run pgstream between AWS (eu-central-1) and another cloud. We measured this on the target:

| Item | Value |
|---|---|
| Transactions | 220 each second |
| Rows in one transaction | approximately 7 |
| CPU | 1 to 2.5 cores of 4 |
| Backends that wait | 0 |
| Time in WAL fsync | 0 |

The database is idle. The network controls the speed.

The batch size settings cannot correct this. A source keepalive also closes a batch (`wal_batch_sender.go`, "max batch size reached or keep alive received"). Therefore a batch is usually much smaller than `maxBatchSize` or `maxBatchBytes`, and an increase of those limits changes nothing.

## The change

`Tx.ExecBatch` puts the queries into a `pgx.Batch`, sends them together, and then reads the results. `execQueries` uses it in place of the loop.

## The behaviour does not change

The results come back in the same sequence as the queries. Thus the first error identifies the query that failed. Postgres stops the transaction at that query, so all later queries also fail. For this reason `ExecBatch` stops at the first error and returns its index.

`execQueries` then drops the same query as before and retries the same list. Strict mode and the dropped query counter are not different.

The retry path also becomes cheaper. Before this change, k bad queries in a batch cost k+1 runs of N round trips each. Now they cost k+1 exchanges.

## Tests

- All tests in `internal/postgres`, `pkg/wal/processor/postgres` and `pkg/wal/processor/batch` pass without changes. The mock uses `ExecFn` when a test does not set `ExecBatchFn`, so the tests that count calls to `Exec` continue to see one call for each query.
- `TestBatchWriter_execQueries_pipelined` is new. It shows that the whole batch arrives in one call, and that a failure in the middle drops the correct query and retries the others.

## Points for the maintainers

1. **Memory.** The pipeline holds all queries in a buffer before it reads the first result. `maxBatchBytes` limits the size, but the buffer is now on the client and on the server at the same time.
2. **Option.** The change is unconditional. Tell me if you prefer an option for it.
3. **Numbers.** I did not add a benchmark. The histograms of pgstream cannot show this effect: `pgstream_target_processing_latency_nanoseconds` records nanoseconds into the default OTel buckets, and the largest finite boundary is 10000. Therefore every measurement stops at 10 microseconds. I can send a separate pull request for the buckets if that is useful.

### DIFF
--- a/internal/postgres/pg_tx.go
+++ b/internal/postgres/pg_tx.go
@@ -17,6 +17,10 @@ type Tx interface {
 	Query(ctx context.Context, query string, args ...any) (Rows, error)
 	QueryRow(ctx context.Context, dest []any, query string, args ...any) error
 	Exec(ctx context.Context, query string, args ...any) (CommandTag, error)
+	// ExecBatch runs the queries as one pipelined exchange. It does not use one
+	// round trip for each query. It returns the index of the first query that
+	// failed. If all queries are successful, it returns the number of queries.
+	ExecBatch(ctx context.Context, queries []BatchQuery) (int, error)
 	CopyFrom(ctx context.Context, tableName string, columnNames []string, srcRows [][]any) (int64, error)
 	CopyFromText(ctx context.Context, tableName string, columnNames []string, srcRows [][]any) (int64, error)
 	CopyToWriter(ctx context.Context, w io.Writer, sql string) (int64, error)
@@ -61,6 +65,38 @@ func (t *Txn) Query(ctx context.Context, query string, args ...any) (Rows, error
 func (t *Txn) Exec(ctx context.Context, query string, args ...any) (CommandTag, error) {
 	tag, err := t.Tx.Exec(ctx, query, args...)
 	return CommandTag{tag}, MapError(err)
+}
+
+// BatchQuery is one query for ExecBatch.
+type BatchQuery struct {
+	SQL  string
+	Args []any
+}
+
+// ExecBatch sends all queries together and then reads the results. This
+// removes one network round trip for each query. The saving is large when the
+// database is far from the client.
+//
+// The results come back in the same sequence as the queries. Thus the first
+// error identifies the query that failed. Postgres stops the transaction at
+// that query, so all later queries also fail. For this reason ExecBatch stops
+// at the first error.
+func (t *Txn) ExecBatch(ctx context.Context, queries []BatchQuery) (int, error) {
+	batch := &pgx.Batch{}
+	for _, q := range queries {
+		batch.Queue(q.SQL, q.Args...)
+	}
+
+	results := t.Tx.SendBatch(ctx, batch)
+	for i := range queries {
+		if _, err := results.Exec(); err != nil {
+			// Close reports the same failure again. The error from the query is
+			// the useful one, because it has the index.
+			_ = results.Close()
+			return i, MapError(err)
+		}
+	}
+	return len(queries), MapError(results.Close())
 }
 
 // CopyFrom uses pgx's binary-format COPY, which is the fast path for any
--- a/internal/postgres/pg_tx.go
+++ b/internal/postgres/pg_tx.go
@@ -87,7 +87,7 @@ func (t *Txn) ExecBatch(ctx context.Context, queries []BatchQuery) (int, error) 
 		batch.Queue(q.SQL, q.Args...)
 	}
 
-	results := t.Tx.SendBatch(ctx, batch)
+	results := t.SendBatch(ctx, batch)
 	for i := range queries {
 		if _, err := results.Exec(); err != nil {
 			// Close reports the same failure again. The error from the query is
--- a/pkg/wal/processor/postgres/postgres_batch_writer.go
+++ b/pkg/wal/processor/postgres/postgres_batch_writer.go
@@ -319,18 +319,25 @@ func (w *BatchWriter) execQueries(ctx context.Context, queries []*query) ([]*que
 			return err
 		}
 
-		for i, q := range queries {
-			if _, err := tx.Exec(ctx, q.sql, q.args...); err != nil {
-				w.logger.Error(err, "executing sql query", loglib.Fields{
-					"sql":  q.sql,
-					"args": q.args,
-				})
-				// if a query returns an error, it will abort the tx. Remove it
-				// from the list of queries to be retried.
-				droppedQuery = q
-				retryQueries = removeIndex(queries, i)
-				return err
-			}
+		// Send the queries as one pipeline. One round trip for the batch
+		// replaces one round trip for each query. The behaviour does not
+		// change: ExecBatch returns the index of the query that failed, which
+		// is the same index the loop found before.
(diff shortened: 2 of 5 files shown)


pkg/wal/processor/postgres/postgres_batch_writer.go
```diff
@@ -319,18 +319,25 @@ func (w *BatchWriter) execQueries(ctx context.Context, queries []*query) ([]*que
...
+		// Send the queries as one pipeline. One round trip for the batch
+		// replaces one round trip for each query. The behaviour does not
+		// change: ExecBatch returns the index of the query that failed, which
+		// is the same index the loop found before.
+		batchQueries := make([]pglib.BatchQuery, 0, len(queries))
+		for _, q := range queries {
+			batchQueries = append(batchQueries, pglib.BatchQuery{SQL: q.sql, Args: q.args})
+		}
+
+		if i, err := tx.ExecBatch(ctx, batchQueries); err != nil {
```
kvch, line 331: It is possible that `results.Close()` returns an error after we processed all queries. So if we executed 5 queries and err is not nil, `ExecBatch` returns `5, err`. But the length of `queries` is 5, so `queries[i]` panics with index out of range.

internal/postgres/pg_tx.go
```diff
@@ -63,6 +67,38 @@ func (t *Txn) Exec(ctx context.Context, query string, args ...any) (CommandTag,
...
+// at the first error.
+func (t *Txn) ExecBatch(ctx context.Context, queries []BatchQuery) (int, error) {
+	batch := &pgx.Batch{}
+	for _, q := range queries {
+		batch.Queue(q.SQL, q.Args...)
+	}
+
+	results := t.SendBatch(ctx, batch)
+	for i := range queries {
+		if _, err := results.Exec(); err != nil {
```
kvch, line 92: It is possible that `SendBatch` fails before any of the queries are executed. In this case the code mistakenly treat `queries[0]` as the query that is failing. Please handle batch errors separately for actual query errors and for other issues.

kvch (changes requested):
Thank you for the contributions. 2 notes must be addressed before merging this PR.

pkg/wal/processor/postgres/postgres_batch_writer.go
siriusfreak: Fixed in 44d1852.

`ExecBatch` now returns an `error` rather than an index, so the out-of-range read is no longer expressible. A query the server rejects comes back as a `*BatchQueryError` carrying its index; `results.Close()` failing after every query has answered comes back as a plain error, and the writer no longer looks for a query to blame in that case.

Worth noting that the panic you spotted is on the send path, where nothing recovers: the `recover()` in the batch writer covers `ProcessWALEvent`, not `execQueries`. So this was a process exit, not a logged error.

internal/postgres/pg_tx.go
siriusfreak: Fixed in 44d1852, and thank you for making the distinction explicit, because my first attempt at this did not hold. Wrapping the first read failure with index 0 keeps the same wrong answer, just in a nicer type: a send that never reaches the server still fails on the first `results.Exec()`.

The discriminator is who answered. Only a `*pgconn.PgError` means the server rejected that query, so that case becomes a `BatchQueryError` with the index. Anything else (transport, context, a connection that went away) is the batch failing around the queries and comes back plain.

The writer then drops a query only when one is named. A batch failure with nothing to blame travels up instead of returning an empty retry list, which previously dropped the whole batch silently.

`TestBatchQueryError` covers the wrapper, and `TestBatchWriter_execQueries_pipelined` gained a case for a batch failure with no query to blame.

siriusfreak:
Pushed a correction to the attribution itself.

The version this pull request carried still blamed the wrong query in a case that matters. pgx prepares every statement of the batch, and encodes every argument, before it executes any of them, so a failure of either step surfaces on the **first** read whichever statement owns it. `errors.As` finds a `*pgconn.PgError` there, and the index of that first read is not its owner. Query 0 valid and query 1 against a missing table was enough to drop query 0 and let query 1 run again — the loss this attribution is meant to prevent.

What changed:

- `pgx.ErrPreprocessingBatch` carries the SQL of its statement, so the failure is now matched back to a position. When two queries of the batch share that SQL the match is not unique and the failure stays with the batch rather than guessing.
- A failure that names no query no longer costs the whole batch. The writer runs a second pass that executes the queries one at a time inside a transaction it rolls back, so the server names the one at fault and nothing is applied. One query is dropped again rather than all of them, which matters most with `ignore_send_errors`.
- The index is checked before it reaches the slice, since an out-of-range index on the send path is the panic this work started from.

`TestBatchError` covers a preprocessing failure naming a later query, the ambiguous case where the SQL repeats, a server error for the query just read, a failure that is not the server, and an out-of-range index. The preprocessing error is a stand-in type because `pgx.ErrPreprocessingBatch` has unexported fields; the code matches on the shape that names a statement, and the real type satisfies it.

Found by a k3s harness for this pipeline: it logged an error about `a_vchar`, a column of one table, next to an `INSERT` into another table that has no such column.

siriusfreak:
Pushed 2d315a0, which fixes a defect in the isolation pass I added earlier in this PR.

The pass runs the queries of a failed batch one at a time inside a transaction it rolls back, and it asked for that rollback by returning a private sentinel error. `retrier.IsRetriableError` keeps anything it does not recognise, so the sentinel qualified: the retrier ran the whole pass again, calling `resetConn` between rounds, until the elapsed budget ran out — and a constant retry policy has no budget, so there it never stopped. The pass always had its answer on the first try.

A rollback the caller asks for is now `postgres.ErrTxRollback`, and `IsRetriableError` treats it as final, so any transaction that ends this way runs once. `TestQuerier_ExecInTx_requestedRollbackRunsOnce` holds it with the real retrier, and fails the test if `connBuilder` is called at all.

Two smaller things in the same commit:

- The pass returns the error the isolated query failed with, not just its index, and `execQueries` records that one. A dropped query now carries the sentence naming the column or the value rather than the batch wrapper.
- `pg_tx_test.go` asserts the `SQL() string` shape against `pgx.ErrPreprocessingBatch` itself. The stand-in satisfied that interface by construction and proved nothing about pgx, so a release that stopped reporting the SQL would have turned the attribution branch into dead code silently.

It stops at the first rejected query on purpose: a statement the server rejects aborts its transaction and everything after it fails with `25P02` whatever it says, so naming a second one in the same pass would need a savepoint around every query.

kvch (approved):

PR #1206, opened 2026-09-15
Unquote identifiers correctly before COPY

siriusfreak:
## What happens

A column whose name holds a double quote stops the snapshot for the whole
database:

```
ERR failed to send batch  error.message="copy from: relation does not exist: column \"a \"\"b\" does not exist"  module=postgres_bulk_ingest_writer table=quoted_column
ERR snapshot generation failed  error.message="copy from: relation does not exist: column \"a \"\"b\" does not exist"
ERR pg snapshot and listen      error.message="copy from: relation does not exist: column \"a \"\"b\" does not exist"
```

The listener then exits. One column name takes a standby down, and the name in
the message is not the name of the column.

## Why

`removeQuotes` is `strings.Trim(s, "\"")`. A name that holds a double quote
carries it doubled inside the quoted form, so `"a ""b"""` loses one quote from
the front and three from the back and becomes `a ""b`. pgx quotes what it is
given, and postgres sees `"a ""b"`.

`UnquoteIdentifier` in the same file already reverses the quoting exactly, and
its doc comment describes this case. It was simply not used here.

## What changed

- `removeQuotes` delegates to `UnquoteIdentifier`.
- The four COPY call sites use a new `unquoteIdentifiers`, which returns a new
  slice. The old loop wrote through the caller's slice, so after one COPY the
  query held unquoted names and a retry built different SQL.
- Two unit tests: the embedded-quote case, and a round trip of
  `UnquoteIdentifier(QuoteRawIdentifier(name))`.

## How to verify

```sql
CREATE TABLE quoted_column ("id" integer PRIMARY KEY, "a ""b""" text);
INSERT INTO quoted_column VALUES (1, 'first');
```

Snapshot that table into an empty target. Before this change the snapshot
fails with the error above. After it the row arrives.

`go test ./internal/postgres/...` covers the unquoting itself.

## Risk

Low. The change makes an existing helper the single way identifiers are
unquoted, and the new slice removes a write through the caller's memory.


internal/postgres/pg_utils.go
```diff
@@ -159,8 +159,26 @@ func newIdentifier(tableName string) (pgx.Identifier, error) {
 	return identifier, nil
 }
 
+// removeQuotes reverses the quoting of one identifier.
+//
+// It has to reverse it exactly. A name that holds a double quote carries it
+// doubled inside the quoted form, so a trim of the outer quotes leaves the
+// doubling behind. pgx quotes what it is given, and postgres then answers
+// `column "a ""b" does not exist` and the whole COPY fails.
 func removeQuotes(s string) string {
```
kvch, line 168: `removeQuotes` is no longer used, please remove it along with the long comment. You can just use `UnquoteIdentifier` where necessary.

kvch:
Small comment, otherwise look good

internal/postgres/pg_utils.go
siriusfreak: Done in 611602e3 — the wrapper and its comment are gone, and `newIdentifier` calls `UnquoteIdentifier` directly.

You were right that the comment was the wrong place for that argument: the commit message for the original change already said "UnquoteIdentifier already reverses the quoting exactly, use it", and then the code added a name to read through anyway. The exactness is documented on `UnquoteIdentifier` itself, `"my""table"` -> `my"table` among its examples, which is where someone would look for it.

kvch (approved):

PR #1207, opened 2026-09-15
Keep the digits of a numeric on the replication path

siriusfreak:
## What happens

A `numeric` loses its digits over the replication path. The load path keeps the
value, so a fresh replica looks correct and then diverges from the first write
onward. Nothing reports it: no error, no dropped query, and the slot stays
active with no lag.

Measured on a source and a target of the same version, one row at a time:

| on the source | on the target |
| --- | --- |
| `106.000000000000000001` | `106` |
| `123456789012345678901234567890.123456789` | `123456789012345680000000000000` |
| `0.10000000000000000000000000001` | `0.1` |
| `-9223372036854775808.5` | `-9223372036854776000` |

The last row is the one to look at: a value next to the bigint bound comes back
as a different integer.

## Why

wal2json prints every postgres number as a JSON number. The listener decodes
with `UnmarshalUseInt64`, whose comment already names the gap: "Non-integer
numbers still decode as float64." A float64 holds about 17 significant digits
and a numeric holds up to 1000.

This is the same shape as the bigint rounding that `UseInt64` was added to
repair. Only the integer half was repaired.

## What changed

- `internal/json` gains `UnmarshalUseNumber`, which decodes every number into
  `json.Number` and keeps the text the source printed.
- The listener uses it, and a new pass gives each value the Go type its column
  needs: an `int64` for an integer column, a `float64` for a float column, and
  the text itself for everything else, which postgres parses with no loss.
- The pass sits next to `decodeByteaColumns`, for the reason its comment gives:
  this is the only place that knows the value came from wal2json.
- Unit tests for each type and for the whole pass.

Integer columns keep the behaviour `UseInt64` gave them, including values above
2^53.

## How to verify

```sql
CREATE TABLE precise_numbers (id integer PRIMARY KEY, v numeric, w double precision);
INSERT INTO precise_numbers VALUES (1, '106.000000000000000001', 0.1);
```

Replicate that row through the WAL, not the snapshot, then compare. Before this
change the target holds `106`.

`go test ./pkg/wal/listener/postgres/... ./internal/json/...`

## Risk

Medium, because it touches the type of every value on the hot path. The
integer and float cases are pinned by tests, and everything else arrives as
text, which is what postgres reads anyway.


siriusfreak:
Reworked after review: the identities do not move.

The first version changed the value for every sink, and an identity built from it moved with it. For a search index with a `numeric(10,2)` primary key of `1.00`, `parseIDColumns` used to receive `float64(1)` and produce `<tableID>_1`; it would now receive `"1.00"` and produce `<tableID>_1.00`. After an upgrade an update writes a second document and a delete removes neither, so the index keeps stale rows. A Kafka topic keyed by primary key moves its message keys for the same reason, and with them the partition a row lands in.

`Column.IdentityValue` now renders a numeric the way it was rendered before, while the column value keeps every digit. The search adapter and the Kafka key builder use it, and the rest of the change is unaffected.

A numeric that a float64 cannot hold had a colliding identity before this change and still has one. That is a separate question from the value, and this pull request does not change it.

pkg/wal/processor/kafka/wal_kafka_batch_writer.go
```diff
@@ -311,7 +311,9 @@ func primaryKeyMessageKey(walData *wal.Data) []byte {
 		if !found {
 			return nil
 		}
-		values = append(values, keyValueEscaper.Replace(fmt.Sprintf("%v", col.Value)))
+		// IdentityValue keeps the rendering a previous version produced for a
+		// numeric column, so that the key of an existing topic does not move.
+		values = append(values, keyValueEscaper.Replace(fmt.Sprintf("%v", col.IdentityValue())))
```
kvch, line 316: The key does not keep its previous rendering.

The old decoder (`UnmarshalUseInt64`) turned *any* integral JSON number into `int64`, whatever the column type, so a `numeric` id of `1234567` rendered as `1234567`. `IdentityValue` always returns `float64`, and `fmt.Sprintf("%v", float64)` uses `%g`.

The same applies to `double precision` and `real` identity columns holding integral values: `numberValue` now decodes them as `float64` where the old decoder gave `int64`, and `IdentityValue` does not touch them at all, so their keys move too.

Consequence after an upgrade: the key for an unchanged row moves, so it hashes to a different partition (per-row ordering breaks across the upgrade), and log compaction no longer collapses the old and new versions of that row — the breakage this commit set out to avoid.

Rendering the identity through `strconv.FormatFloat(v, 'f', -1, 64)` (or keeping an integral value as an integer) would reproduce the old output.

kvch (commented):
## Review: numeric precision on the replication path

The direction is right and the postgres target handles the text everywhere, including the coalesced `DELETE ... = ANY($1::numeric[])` path (pgx retries the other wire format when the preferred one has no encode plan — confirmed with a real DELETE on a `numeric` primary key). Binary COPY is reached only by the bulk ingest writer, which is snapshot-only, so it never sees these strings. `decodeNumberColumns` is nil-safe, leaves no `json.Number` in the event, and covers both `Columns` and `Identity`. The `integerTypes`/`floatTypes` sets match exactly the types wal2json format-version 2 prints unquoted; `NaN`/`Infinity` arrive quoted, and arrays, `json`/`jsonb` and domains arrive as strings as before.

Three findings need a fix before merge. Two of them are silent at upgrade: no error is logged and no test fails, but identities move and collide.

1. **Kafka message keys change shape for every numeric and float identity column.** `IdentityValue` always returns `float64`, and `fmt.Sprintf("%v", float64)` uses `%g`.
2. **Search document ids lose precision above 2^53 and can collide.** The comment saying this collision pre-exists is not correct.
3. **`greenmask_integer` stops working on `numeric` columns**, with a pass-through path that writes the un-anonymised value.

Details are in the inline comments, with the empirical check for each.

---
🤖 Review produced with [Claude Code](https://claude.com/claude-code)

pkg/wal/wal_data.go
```diff
@@ -102,3 +104,42 @@ func (m Metadata) IsIDColumn(colID string) bool {
...
+// search document id, or a kafka message key.
+//
+// A numeric column carries its text, so that its digits survive the trip. That
+// text is not what a previous version produced here. It built an identity from
+// the float64 its decoder gave it, so a numeric(10,2) of 1.00 became "1". An
+// index and a topic written by that version must keep matching after an
+// upgrade, or an update writes a second document and a delete removes neither.
+//
+// The identity therefore keeps the old rendering, while the column value keeps
+// every digit. A numeric that a float64 cannot hold had a colliding identity
```
kvch, line 118: This claim does not hold, and the ids move at upgrade.

Before this change, an integral `numeric` up to 2^63 decoded to an exact `int64`, so `9007199254740993` produced document id `..._9007199254740993`. It now goes text -> `float64` -> `9007199254740992`. Two consequences:

- The id of every row with a `numeric` key above 2^53 changes at upgrade. An UPDATE then writes a second document, and the DELETE removes only one of the two.
- `9007199254740992` and `9007199254740993` collapse onto the same document id, so one row silently overwrites the other.

A `numeric(20,0)` primary key — the usual shape after an Oracle or snowflake-id migration — hits this on every row, so it is not a corner case.

If the identity has to keep the old rendering, it needs to keep the old *type* too: parse the text as an integer first and only fall back to `float64` when it has a fractional part.

pkg/transformers/greenmask/greenmask_float_transformer.go
```diff
@@ -107,6 +108,16 @@ func (ft *FloatTransformer) Transform(_ context.Context, value transformers.Valu
 		toTransform = getBytesForFloat(float64(val))
 	case int:
 		toTransform = getBytesForFloat(float64(val))
+	// A numeric column carries its text on the replication path, so that a
+	// value wider than a float64 keeps its digits on the way to the target.
+	// The transformer produces a float64 whatever it is given, so the text is
+	// read as one here, as it was before it was text.
+	case string:
```
kvch, line 115: Oonly the float transformer learned this case, but `greenmask_integer` also accepts `numeric`.

`docs/transformers.md:501` lists `numeric` as a supported type for `greenmask_integer`, and `pgTypeCompatibleWithTransformerType` accepts it (NumericOID -> `Float64DataType`). `IntegerTransformer.Transform` (`greenmask_integer_transformer.go:120`) has no `case string`.


So a rule that worked before this change either stops the stream (`on_error: fail`) or, with `on_error: pass-through`, logs a warning and writes the **original un-anonymised value** to the target — the path logged with `severity: DATALOSS`.

`greenmask_unix_timestamp` embeds `IntegerTransformer` and inherits the same break. Every transformer that declares `numeric` support needs the string case, not just this one.

pkg/transformers/greenmask/greenmask_float_transformer.go
```diff
@@ -107,6 +108,16 @@ func (ft *FloatTransformer) Transform(_ context.Context, value transformers.Valu
...
+		number, parseErr := strconv.ParseFloat(val, 64)
+		if parseErr != nil {
+			return nil, transformers.ErrUnsupportedValueType
```
kvch, line 118: A range error is fatal here.

`float64FromNumeric` clamps an out-of-range value, which is what `docs/transformers.md:482` documents: "If the source value is outside the `float64` range, the transformer uses the largest `float64` value." A `numeric` holding `1e400` is legal (numeric reaches ~1e131072), so the same value now fails the transform on the replication path while it still works from a snapshot.

`strconv.ParseFloat` returns `±Inf` together with `strconv.ErrRange`, so accepting the value when `errors.Is(parseErr, strconv.ErrRange)` restores the documented behaviour and keeps the two paths in agreement.

pkg/wal/processor/search/search_adapter.go
```diff
@@ -207,7 +207,9 @@ func (a *adapter) parseIDColumns(tableName string, idColumns []wal.Column, doc *
 
 	isCompositeID := len(idColumns) > 1
 	for _, col := range idColumns {
-		switch v := col.Value.(type) {
+		// IdentityValue keeps the rendering a previous version produced for a
+		// numeric column, so that an existing index keeps matching.
+		switch v := col.IdentityValue().(type) {
```
kvch, line 212: Non-id `numeric` columns now reach the indexer as Go strings.

This switch fixes the id, but the column *value* is unchanged. `numeric` maps to an ES/OpenSearch `double` (`pkg/wal/processor/search/store/search_pg_mapper.go:149`), which coerces a numeric string, so nothing errors — the stored `_source` for that field simply flips from `1.50` to `"1.50"`, while it stays a JSON number on the snapshot path (`pgtype.Numeric.MarshalJSON`).

One index then holds both shapes for the same field, and a client reading `_source` rather than `fields` sees a type change with no error anywhere. Converting the value to a float for the search target, or documenting the change, would close this.

pkg/wal/listener/postgres/number.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+}
+
+// numberValue converts one JSON number.
+//
+// An integer column becomes an int64, which is what the writer and the
+// transformers expect. A float column becomes a float64, because a float64 is
+// what the source holds. Every other numeric column keeps its text, which
+// postgres parses with no loss.
+func numberValue(num json.Number, colType string) any {
```
kvch, line 63: The text reaches every serialising target, not only postgres.

The PR's Risk section and the docs discuss the postgres target, but this decoder sits in the listener, so the text flows into the kafka message value, the webhook payload and the stdout writer as well: they now emit `"value": "1.50"` where they emitted `"value": 1.50`.

That is a breaking change for any external kafka or webhook consumer with a typed schema. pgstream's own kafka reader is unaffected, since it writes the text straight back to postgres.

pkg/wal/listener/postgres/number.go
```diff
@@ -0,0 +1,77 @@
...
+		cols[i].Value = numberValue(num, cols[i].Type)
+	}
+}
+
+// numberValue converts one JSON number.
+//
+// An integer column becomes an int64, which is what the writer and the
+// transformers expect. A float column becomes a float64, because a float64 is
+// what the source holds. Every other numeric column keeps its text, which
+// postgres parses with no loss.
```
kvch: **Low — deterministic transformers keyed on the raw value change their output for `numeric` columns.**

`lookupValueKey` (`pkg/transformers/lookup_choice_transformer.go:398`) hashes `fmt.Sprintf("%v", v)`, so after an upgrade the same row selects a different lookup value than the rows already written with it.

`TemplateTransformer` also receives a string instead of a `float64`. That actually makes it agree with the snapshot path, which is `templateRenderable`'s stated goal, but a template using a float verb (`printf "%.2f" .GetValue`) silently starts rendering `%!f(string=1.50)`.

internal/json/json.go
```diff
@@ -23,6 +23,19 @@ func UnmarshalUseInt64(b []byte, v any) error {
...
+// text the source printed. A float64 holds about 17 digits and a postgres
+// numeric holds up to 1000, so a numeric that goes through a float64 comes
+// back rounded.
+var useNumberAPI = json.Config{UseNumber: true}.Froze()
+
+// UnmarshalUseNumber behaves like Unmarshal but decodes every JSON number into
+// json.Number when the destination is an interface{}. The caller then gives
+// each value the type its column needs.
+func UnmarshalUseNumber(b []byte, v any) error {
+	return useNumberAPI.Unmarshal(b, v)
```
kvch: **Low — no direct test for the new decoder.**

The PR body says "Unit tests for each type and for the whole pass", but `internal/json/json_test.go` gains no case for `UnmarshalUseNumber`. Its only coverage is indirect, through `pkg/wal/listener/postgres`. A case pinning that an integral number, a fractional one and a 40-digit one all come back as `json.Number` with their text intact would guard the decoder itself.

PR #1209, opened 2026-09-15
Keep the dimensions of an array through the snapshot

siriusfreak:
## What happens

An array of more than one dimension loses its shape through the snapshot:

| on the source | on the target |
| --- | --- |
| `{{1,2},{3,4}}` | `{1,2,3,4}` |
| `{{9}}` | `{9}` |
| `{{1},{2}}` | `{1,2}` |

There is no error and no dropped row. The load reports success and the
counters stay at zero, so nothing says the replica differs.

One dimension is not affected: NULL elements, separators, quotes, braces,
unicode and empty arrays all arrive unchanged.

## Why

`rows.Values()` flattens. Measured against PostgreSQL 14, for a column of type
`int[][]` holding `{{1,2},{3,4}}`:

```
rows.Values()                      -> []interface {}  [1 2 3 4]
pgtype.Array[int32] from RawValues -> dims=[{2 1} {2 1}] elements=[1 2 3 4]
```

The dimensions are in the wire data. pgx simply does not carry them into
`Values()`.

Writing the value back needs the same care. Of the three shapes, only one
round trips:

```
rows.Values() flat []interface{}   -> {1,2,3,4}        wrong
pgtype.Array[int32] with dims      -> {{1,2},{3,4}}    ok
nested [][]int32                   -> {1,2,3,4}        wrong
```

Nested Go slices are flattened by pgx as well, so they are not a way out.

## What changed

`rowValues` decodes each array column from the raw bytes into
`pgtype.Array[any]`, and replaces the value only when the array has more than
one dimension. Everything else keeps the type `rows.Values()` gives it, so a
value that already arrives correctly does not change shape for a transformer or
for another target.

The `Rows` mock returns nil rather than panicking when `RawValuesFn` is unset,
which is what the existing tests need.

`pgtype.Array[any]` was checked through `CopyFrom` for every shape that came to
mind: one and two dimensions, `int`, `text` with commas, quotes and NULL,
`numeric`, `timestamptz`, the empty array and the all-NULL array. All of them
round trip.

## How to verify

```sql
CREATE TABLE nested_arrays (id integer PRIMARY KEY, v integer[][]);
INSERT INTO nested_arrays VALUES (1, '{{1,2},{3,4}}');
```

Snapshot that table into an empty target and compare. Before this change the
target holds `{1,2,3,4}`.

## Risk

Low. The replacement is limited to the case that is broken today, and a decode
that fails keeps the old value rather than stopping the load.


siriusfreak:
Reworked after review: the value no longer carries a `pgtype.Array`.

The first version was wrong for every target except postgres. `pgtype.Array` has no JSON form, so a Kafka target writes an object of `Elements`, `Dims` and `Valid`, the reader hands the postgres adapter a `map[string]any`, and the array codec rejects it. A multidimensional row that used to arrive flattened would not have arrived at all, which is a worse failure than the one this fixes.

The snapshot now carries the postgres array literal. It is a plain string, it is what the replication path already carries for every array, it survives JSON, and the postgres sink parses it back into a `pgtype.Array` so the pgx value never leaves that sink.

The literal was checked against what postgres itself prints, for `int[][]`, `text[][]` with commas, quotes, braces and NULL, `int[][][]`, `numeric[][]` and `int[]`: identical in each case. For `timestamptz[][]` pgx writes `2026-01-01 00:00:00Z` where postgres writes `"2026-01-01 00:00:00+00"`, which is the same instant and parses back to the same value.

kvch (commented):
The code looks good to me. Could you please add an automated test for this?

siriusfreak:
Added in 8196226, one test for each half of the change.

`TestRowValues` feeds the wire bytes of `{{1,2},{3,4}}` through `rowValues` and expects the literal. It fails against `rows.Values()` alone, which gives `[1 2 3 4]` — that is the loss this PR is about, so the test bites rather than describing what the code already does. The other cases pin what must not change: one dimension, a column that is not an array, and a NULL all keep the value `rows.Values()` gave, and a reader that carries no wire data returns its rows untouched.

`Test_updateValueForCopy_arrayDimensions` covers the other end, where the sink reads the literal back for binary COPY. It asserts the dimensions and not only the elements, because that is the part `pgtype.FlatArray` silently dropped; against `FlatArray` it fails.

Both were checked by reverting the change and watching them fail, and `go test ./...` is green.

One thing the unit tests cannot reach, in case it matters to you: neither exercises a real `COPY` into postgres. The end-to-end case runs in a separate harness of ours, against a k3s cluster, where a multidimensional column arrived as `{1,2,3,4}` before this change and matches the source after it. I am happy to add an integration test here too if you would prefer the coverage in the repository.

kvch (approved):

PR #1212, opened 2026-09-17
Fix explicit sequence discovery during CDC

### LINKED ISSUE #1203: Sequence current value not advanced on target during CDC replication
**Describe the bug**

The WAL writer fails to keep sequences synchronised on the target database when sequences are created explicitly (CREATE SEQUENCE) and referenced via DEFAULT nextval(...). After cutover, the target sequence is stuck at its snapshot-time value while row data has caught up, causing the first application INSERT using DEFAULT to collide on an already-existing primary key.

**To Reproduce**

-- On source
CREATE SEQUENCE myschema.sq_test START WITH 1 INCREMENT BY 1;
CREATE TABLE myschema.seq_probe (
    id  bigint PRIMARY KEY DEFAULT nextval('myschema.sq_test'),
    tag text
);
INSERT INTO myschema.seq_probe (tag)
    SELECT 'seed-' || g FROM generate_series(1, 1000) g;

Start pgstream replication. Once CDC is running, insert more rows on source:

INSERT INTO myschema.seq_probe (tag)
    SELECT 'cdc-' || g FROM generate_series(1, 500) g;

Wait for rows to replicate, then stop replication and check target:

SELECT last_value FROM myschema.sq_test;
-- returns 1000 (snapshot-time value, not 1500)

SELECT max(id) FROM myschema.seq_probe;
-- returns 1500 (rows replicated correctly)

INSERT INTO myschema.seq_probe (tag) VALUES ('test');
-- ERROR: duplicate key value violates unique constraint
-- DETAIL: Key (id)=(1001) already exists.

**Expected behavior**
During CDC replication, the target sequence last_value should be advanced to match the source after every replicated INSERT, regardless of how the sequence was created. After replication stops, SELECT last_value FROM myschema.sq_test on the target should equal max(id) in the replicated table, and subsequent INSERTs using DEFAULT nextval() should succeed without primary key collisions.

**Setup (please complete the following information):**
 - pgstream version: 1.3.1
 - Postgres version: 16
 - Postgres environment: AWS RDS

mathewjustin:
#### Description

Discover sequence-backed columns whose sequence was created explicitly and referenced by a direct `DEFAULT nextval(...)`, without `OWNED BY`. This keeps the existing PostgreSQL writer sequence-update path working after the initial snapshot.

##### Related Issue(s)

- Closes #1203

#### Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] Test coverage improvement

#### Changes Made

- Follow the normal `pg_attrdef -> sequence` dependency used by direct `nextval` defaults instead of relying only on the sequence-owned-by-column dependency.
- Require the canonical default expression to be exactly `nextval(sequence::regclass)`, so derived expressions such as `nextval(sequence) + 100` are not treated as direct sequence values.
- Preserve identity-column discovery and return the sequence's actual schema for cross-schema defaults.
- Add a live PostgreSQL catalog regression test for PostgreSQL 14 and 17.

#### Testing

- [x] All existing tests pass

Commands run:

```text
make test
make integration-test
golangci-lint run --timeout=30m --config=.golangci.yml
```

The regression test was also run against unmodified `main`: it returned only the `bigserial` mapping and missed both the identity column and the explicit unowned sequence. The same test passes with this change on PostgreSQL 14 and 17.

#### Checklist

- [x] Self-review completed
- [x] Code is well-commented

#### Additional Notes

This does not change the writer's existing `setval` behavior. It changes only how the schema observer discovers which columns directly represent sequence values.

#### Reproduction Lab

A standalone Docker Compose lab reproduces the failure and runs the same workload with this patch:

https://github.com/mathewjustin/pgstream-sequence-lab

```bash
git clone https://github.com/mathewjustin/pgstream-sequence-lab.git
cd pgstream-sequence-lab
./demo.sh baseline
./demo.sh fixed
# Or run both sequentially:
./demo.sh both
```

Both variants build from pgstream `v1.3.1`; the only difference is whether the candidate patch is applied. After snapshotting 1,000 rows and replicating 500 more through PostgreSQL -> pgstream -> Kafka -> pgstream -> PostgreSQL, the lab observes:

| Variant | Target `max(id)` | Target sequence `last_value` | First target insert |
| --- | ---: | ---: | --- |
| Baseline | 1500 | 1000 | Duplicate key on `id=1001` |
| Fixed | 1500 | 1500 | Succeeds with `id=1501` |


kvch (changes requested):
Could you please also make sure `updateColumnSequences` behaves correctly? It seems to me that we loose the sequence info during updates because identity columns do not have default because they are not in `pg_attrdef`.

